Issue #22, opened 2025-07-21
blog page could use a refresh


linsun:
Hi @artberger not sure if you have some bandwith for this, we will start to leverage the blog page to publish more blogs for agentgw.

Could you help us refresh the blog page and add [this blog](https://www.solo.io/blog/updated-a2a-and-mcp-gateway) as a seeding blog?

https://agentgateway.dev/blog/

Once that is done, we can start add the blog link on the header of the agentgateway.dev landing page.  WDYT?

Thanks!

artberger:
The blog is live! https://agentgateway.dev/blog/

PR #24, opened 2025-07-25
Add blog to nav bar, styling, and seeder blog

artberger:
Addresses https://github.com/agentgateway/website/issues/22
* Adds blog to website navigation
* Updates blog styling for contrast against the backgrounds
* Adds starter blog


assets/css/custom.css
williamgrh, line 1: So on kagent we've done this in the `blog/single.html` layout. You could either throw all this in a `<style>` tag in there or you could do it like it's done here https://github.com/kgateway-dev/kgateway.dev/blob/main/layouts/blog/single.html#L24 where we use Tailwind styles to style the content.

If it seems like too much to get it working with Tailwind I'd be fine just isolating these styles to the blog layout files inside `<style>` tags.

Everything else looks great!

artberger: I removed the overrides in this custom.css file. I had already copied the `blog/single.html` from the kgateway repo, but I tried a different approach inline in that file that seemed to do something similar as the custom css was. I am not sure if that is sort of like what you meant?

williamgrh: Yep! Your approach is exactly what I meant, it's definitely a little more complex than the kgateway use-case. Looks good to me!

PR #67, opened 2025-10-21
Add transformations, extproc, gateway policies and new images

Nadine2016:


content/docs/configuration/traffic-management/extproc.md
```diff
@@ -0,0 +1,90 @@
...
+4. The modified request headers are sent back to agentgateway.
+5. The modified headers are added to the request.
+6. The request is forwarded to the backend service.
+
+## Failure modes
+
+You can choose whether you want agentgateway to forward requests if the external processing server is unavailable with the `extProc.failureMode` setting. Choose between the following modes: 
+
+* **failOpen**: Block requests if the request to the external processing server fails. This is the default behavior. 
+* **failClosed**: Forward requests to the backend service, even if the connection to the external processing server fails.
```
artberger: Is there a hint we could give as to why they might want to do this? maybe something like:
```suggestion
* **failClosed**: Forward requests to the backend service, even if the connection to the external processing server fails. You might choose this option to ensure availability of the backend services even when the ExtProc service is down.
```

content/docs/configuration/traffic-management/transformations.md
```diff
@@ -0,0 +1,115 @@
...
+---
+
+Agentgateway uses transformation templates that are written in Common Expression Language (CEL). CEL is a fast, portable, and safely executable language that goes beyond declarative configurations. CEL lets you develop more complex expressions in a readable, developer-friendly syntax.
+
+To learn more about how to use CEL, refer to the following resources:
+
+- [cel.dev tutorial](https://cel.dev/tutorials/cel-get-started-tutorial)
+- [Agentgateway reference docs](https://agentgateway.dev/docs/reference/cel/)
+
+{{< callout >}}
```
artberger: I am not sure if this is intentional...without the `type` attribute, it shows up as orange with no icon:
image

with the type=info, it looks like this:
image
```suggestion
{{< callout type="info">}}
```

content/docs/configuration/traffic-management/transformations.md
```diff
@@ -0,0 +1,115 @@
...
+transformations:
+  request:
+    body:
+      '"This is a custom request body."'
+  response:
+    body:
+      '"This is a custom response body."'
+```
+
+{{< callout >}}
```
artberger: same comments as above

hugo.yaml
```diff
@@ -60,6 +60,11 @@ menu:
       weight: 5
       params:
         type: search
+    - name: GitHub
```
artberger: lol i just noticed kgateway does not have this

PR #82, opened 2026-01-08
Fix CORS configuration for the Streamable HTTP MCP example

MaYuan-02:
The current docs only mention allowOrigins and allowHeaders for CORS. However, in browser environments, exposeHeaders is also required; otherwise the browser can’t read the session ID from the response headers.

I ran into this issue while using MCP Inspector in direct mode.
image


howardjohn (approved):
thanks! We have a few examples under examples/ in the main repo we may want to update as well.

PR #120, opened 2026-01-28
adds chat interface for docs assistant

rinormaloku:


layouts/partials/chatbot.html
```diff
@@ -0,0 +1,192 @@
+<!-- Hidden mobile menu elements to prevent Hextra theme main.js errors -->
```
artberger: Does this mean it won't be on mobile view? seems like in the preview, it is not unless you scroll right to pull up these hidden menus. 

(Not sure it needs to be in mobile anyways, just thought I'd check.)

artberger (approved):

PR #146, opened 2026-02-11
[docs] Rename local doc set to standalone

artberger:
From several requests, to clarify that agentgateway binary can be run in other environments than just a local development one.


content/docs/standalone/latest/_index.md
```diff
@@ -0,0 +1,10 @@
+---
+linkTitle: "Main (in dev)"
```
Nadine2016: Title I think needs to change

Issue #161, opened 2026-02-13
Search results grouped by section


artberger:
User problem: Search results cover both standalone and k8s variants, which is overall good, but most search results are duplicated in both. So k8s comes before "local" or "standalone" alphabetically. This means if I'm in the standalone docs section and I search for something, I have to scroll past a bunch of k8s results to get to the standalone section. 

Ideas: 
- search only in current section (k8s vs standalone) but have the bottom result be like "See more results in {the other} docs?" 
- Or maybe we could split it into two searches behind the scenes and just order the two searches depending on which page the user is currently on?

PR #169, opened 2026-02-17
[docs] search results by section

artberger:
Addresses #161 

- Search results now default to the section you're in, but at the end, there is a switch button.
- if you're just on the docs landing page and not in a section, it just returns all results

image


Sodman (approved):
This is great, thanks!

PR #175, opened 2026-02-18
Moved the sample app to install, added versions topic, fixed links

kristin-kronstain-brown:
Changes include:
- Moved the sample app from operations to install in Kubernetes
- Added versions topic to Kubernetes
- Fixed links for standalone


assets/agw-docs/pages/reference/versions.md
artberger: This table should be revised: https://kkb-sample-app-loc.agentproxy.pages.dev/docs/kubernetes/latest/reference/versions/

In the agentgateway doc set, we are only using 2.2.x going forward. We also probably want to minimize the references to kgateway. In the first paragraph control plane sentence is probably fine, but in this table, I would change it here and elsewhere on the page.

PR #367, opened 2026-04-07
Added doc tests

kristin-kronstain-brown:


.claude/skills/doc-tests/SKILL.md
```diff
@@ -87,6 +87,19 @@ test: skip
...
+| Conceptual/about pages (pure prose, no code) | No code to run | Add a minimal "try it" example at the bottom that demonstrates the concept with a working config + assertion. This is a content strategy decision — if the page's purpose is purely explanatory, `test: skip` is appropriate | Content decision |
+| Index/hub pages (`_index.md` with card links) | No code, only navigation | These are legitimately `test: skip`. The only possible validation is confirming card link targets exist, which would be a separate link-checking CI job (not a doc test) | Script change (separate job) |
+
+**Key principle:** The existing framework already supports every content type except pure link validation. Most "non-runnable" pages can become testable through content changes alone — adding hidden `{{< doc-test >}}` blocks that write configs, start services in the background, and run lightweight assertions. The question is usually whether the test is worth the CI time, not whether the framework can handle it.
```
artberger: > The question is usually whether the test is worth the CI time, not whether the framework can handle it.

I am not sure about this sentence? I think the question should be whether the page would be expected to be tested or not. So the index pages, no, but the other pages yes, we just have not prioritized it yet.

artberger: I am also wondering about the order of this table, why start a testing skill talking about skip? 

And also wondering if the table could be shortened, it is quite long. I think in general shorter skills perform better (but I could be wrong).

Maybe putting the actual file paths to tutorials and integrations would help claude find the content quicker

assets/agw-docs/pages/agentgateway/integrations/llm-clients-k8s/claude-code.md
```diff
@@ -20,6 +20,10 @@ Create a secret, backend, and route to proxy Claude Code traffic through agentga
    export ANTHROPIC_API_KEY="sk-ant-your-key-here"
    ```
 
+{{< doc-test paths="claude-code-k8s" >}}
+if [ -z "${ANTHROPIC_API_KEY+x}" ]; then export ANTHROPIC_API_KEY="sk-ant-placeholder"; fi
```
artberger: what is this doing? The step already uses a placeholder value so not sure why this is needed?

assets/agw-docs/pages/agentgateway/llm/transformations.md
```diff
@@ -134,6 +174,49 @@ Parse the `model` field from the incoming request body and the upstream response
...
+  namespace: agentgateway-system
+spec:
+  parentRefs:
+    - name: agentgateway-proxy
+      namespace: agentgateway-system
+  rules:
+    - matches:
+        - path:
+            type: PathPrefix
+            value: /openai
```
artberger: why are we using `openai` here and then changing the `v1/chat/completions` to `openai` later? I believe in general we are moving to try to just use the OpenAI standard v1/chat/completions vs rewriting it to openai. Unless this is the point of this transformation topic, to transform the route to /openai?

artberger (approved):

.claude/skills/doc-tests/SKILL.md
kristin-kronstain-brown: Removed tutorial and integrations from the table. They weren't really general enough anyway. Mostly the considerations there are just because we haven't implemented things there yet.

assets/agw-docs/pages/agentgateway/integrations/llm-clients-k8s/claude-code.md
kristin-kronstain-brown: Fair. I think it's checking to see if the variable is already set, if not, using the placeholder. We probably aren't even going to be setting a real api key in here.

assets/agw-docs/pages/agentgateway/llm/transformations.md
kristin-kronstain-brown: You're right, that breaks it.

kristin-kronstain-brown: Fixed

PR #382, opened 2026-04-11
Docs/prereq for inference

AdminTurnedDevOps:
I added the inference for our inference section of the docs:
1. Prerequisite section setting the inference extension to be enabled if someone wants to use inference.
2. A quickstart for setting up inference routing and inference pools


artberger (approved):
Thanks for adding this! I will approve, merge it in, and then run a style pass on it.

PR #415, opened 2026-04-20
docs: Fix docker-compose examples' image attribute

sladkani:


artberger (approved):
thanks for your contribution!

PR #443, opened 2026-04-30
docs: add kserve integration guide

Lasse4:
This PR is part of https://github.com/kgateway-dev/kgateway.dev/issues/606


assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,379 @@
+[KServe](https://kserve.github.io/website/) is a Kubernetes-native model inference platform. By fronting KServe with agentgateway, you can apply agent-aware policies, including token-based rate limiting, to your model serving endpoints without modifying your inference services.
```
npolshakova: We might want to rephrase the `agent-aware` policy to be more specific to policies that can be applied when routing to inference models:


```suggestion
[KServe](https://kserve.github.io/website/) is a Kubernetes-native platform for serving machine learning models. With agentgateway in front of KServe, you can enforce traffic management policies, such as token-based rate limiting, for inference requests without modifying your inference services.
```

assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,379 @@
...
+
+## Before you begin
+
+{{< callout type="info" >}}
+Make sure you installed the Experimental Version.
```
npolshakova: I think we want `Latest` Version here. Experimental is a little confusing since it refers to experimental apis in Gateway API. Maybe rephrase to be version v1.2+ ?

assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,379 @@
...
+     --set kserve.controller.gateway.disableIstioVirtualHost=true \
+     --set kserve.controller.gateway.disableIngressCreation=false \
+     --set kserve.controller.knativeAddressableResolver.enabled=false \
+     --set kserve.controller.gateway.localGateway.gateway="" \
+     --set kserve.controller.gateway.localGateway.gatewayService=""
+   ```
+
+## Step 4: Deploy a mocked LLM with httpbun
+
+Instead of a real model, this guide uses [httpbun](https://httpbun.com/) to serve a mock OpenAI compatible endpoint. httpbun's `/llm/chat/completions` path returns a properly structured OpenAI chat completion response, including `usage.total_tokens` in the response body, which agentgateway reads to enforce token-based rate limits.
```
npolshakova: Do we want to try https://kserve.github.io/website/docs/getting-started/genai-first-llmisvc?_highlight=inferenceservice instead of httpbun here?

assets/agw-docs/pages/integrations/kserve.md
artberger: I am thinking we can remove this note because this will only show up in main, which is for the 1.2 version.

assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,379 @@
...
+
+{{< callout type="info" >}}
+Make sure you installed the Experimental Version.
+{{< /callout >}}
+
+{{< reuse "agw-docs/snippets/prereq.md" >}}
+
+## Step 1: Install cert-manager
+
+1. KServe requires cert-manager for webhook certificates.
```
artberger: Start steps with imperative verbs.
```suggestion
1. Install cert-manager, which KServe requires for webhook certificates.
```

assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,379 @@
...
+             requests:
+               cpu: "100m"
+               memory: "128Mi"
+             limits:
+               cpu: "500m"
+               memory: "256Mi"
+   EOF
+   ```
+
+   Wait for the `InferenceService` to become ready.
```
artberger: Use step formatting for these subsequent checks.
```suggestion
3. Wait for the `InferenceService` to become ready.
```

assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,379 @@
...
+   
+   ```shell
+   kubectl get inferenceservices mock-llm -n kserve-test --watch
+   ```
+   
+   Once `READY` is `True`, KServe creates an `HTTPRoute` that attaches to the agentgateway. Verify it. The route attaches to `kserve/kserve-ingress-gateway` with hostname `mock-llm-kserve-test.example.com`.
```
artberger:
```suggestion
4. Verify that KServe created an HTTPRoute after the Gateway becomes `READY`. The route attaches to `kserve/kserve-ingress-gateway` with hostname `mock-llm-kserve-test.example.com`.
```

assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,379 @@
...
+   
+   ```shell
+   kubectl get httproute mock-llm -n kserve-test -o yaml
+   ```
+
+## Step 5: Create an AgentgatewayBackend
+
+KServe generates the `HTTPRoute` with a plain Kubernetes `Service` as the `backendRef`. Agentgateway only applies token-based rate limiting to traffic that flows through an `AgentgatewayBackend` with `spec.ai.provider` configured, because that is what signals to the proxy that the backend is an LLM and that response bodies contain a `usage.total_tokens` field to count against the rate limit bucket.
```
artberger: The second sentence is a bit complex, maybe split it up and rephrase for readability.

Also, combine with the note later on so you can remove it there, since they say similar things. In this rephrase, I also would remove the language about agentgateway supporting responses natively in AgentgatewayPolicy because we typically do not commit to or imply support for future releases in the docs.
```suggestion
KServe generates the `HTTPRoute` with a plain Kubernetes `Service` as the `backendRef`. However, to apply a token-based rate limiting policy, agentgateway needs the backend to be an AgentgatewayBackend. This way, agentgateway knows that the backend is an LLM that has a response body with the `usage.total_tokens` field to count against the rate limit bucket. In the following steps, you create an AgentgatewayBackend and a second HTTPRoute to route to it as a workaround to the KServe-created, Service-based setup.
```

assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,379 @@
...
+
+1. Create an `AgentgatewayBackend` that points at the httpbun predictor service.
```
artberger: I am not sure what predictor means, is it needed? maybe just 
```suggestion
1. Create an `AgentgatewayBackend` that points at the httpbun service.
```

assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,379 @@
...
+           - name: mock-llm-backend
+             namespace: kserve-test
+             group: agentgateway.dev
+             kind: AgentgatewayBackend
+   EOF
+   ```
+
+   {{< callout type="info" >}}
+   This extra `HTTPRoute` is a current workaround. Token-based rate limiting requires traffic to flow through an    `AgentgatewayBackend` so the proxy knows to inspect the response body for `usage.total_tokens`. A future    agentgateway release may support activating LLM-aware response parsing directly on an `AgentgatewayPolicy`,    which would remove the need for this step.
+   {{< /callout >}}
```
artberger: Maybe move this to the shortdesc and delete here, see previous comment.
```suggestion
```

assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,379 @@
...
+
+## Step 6: Apply token-based rate limiting
+How token counting works: Agentgateway reads `usage.total_tokens` from the JSON response body returned by the inference service. Each request deducts that many tokens from the bucket. When the bucket empties, subsequent requests receive `429 Too Many Requests` until the next fill interval.
+
+1. Apply an `AgentgatewayPolicy` that caps requests at **100 tokens per minute**. The policy targets the `mock-llm-ai` route that flows through the `AgentgatewayBackend`.
```
artberger:
```suggestion
1. Apply an `AgentgatewayPolicy` that caps requests at **100 tokens per minute**. The policy targets the `mock-llm-ai` route that selects the `AgentgatewayBackend`.
```

assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,379 @@
...
+   
+   ```shell
+   kubectl get inferenceservices mock-llm -n kserve-test --watch
+   ```
+   
+   Once `READY` is `True`, KServe creates an `HTTPRoute` that attaches to the agentgateway. Verify it. The route attaches to `kserve/kserve-ingress-gateway` with hostname `mock-llm-kserve-test.example.com`.
+   
+   ```shell
+   kubectl get httproute mock-llm -n kserve-test -o yaml
+   ```
```
npolshakova, line 148: I wonder if it's more confusing if we highlight this route is created by kserve but then don't use it? I think there's a couple options here:
- Remove the call out to ` kubectl get httproute mock-llm -n kserve-test -o yaml` section and only focus on the new HTTPRoute that references the AgentgatewayBackend 
- Add an example policy that you can attach to the HTTPRoute that's generated by KServe (maybe tracing or a transformation?)

npolshakova: For the transformation it could be something like this:
```
  traffic:
    transformation:
      response:
        set:
        - name: x-requested-model
          value: 'string(json(request.body).model)'
        - name: x-actual-model
          value: 'string(json(response.body).model)'
```

danehans:
xref https://github.com/kserve/kserve/issues/5474

danehans:
xref: https://github.com/kserve/kserve/issues/5518

cc: @Lasse4 @npolshakova @howardjohn

assets/agw-docs/pages/integrations/kserve.md
danehans: ^ would be consistent with the current agw i[nference routing docs](https://agentgateway.dev/docs/kubernetes/latest/inference/). However, vLLM CPU has [deps](https://docs.vllm.ai/en/stable/getting_started/installation/cpu/) that can create a roadblock for getting new users started. I suggest using the l[lm-d inference simulator](https://github.com/llm-d/llm-d-inference-sim) for both guides for consistency and since both projects (kserve and agtw) are converging on llm-d.

assets/agw-docs/pages/integrations/kserve.md
Lasse4: We can delete it to make the doc easier to understand.

assets/agw-docs/pages/integrations/kserve.md
Lasse4: alright, deleted

assets/agw-docs/pages/integrations/kserve.md
danehans: +1 on the ^ feedback from @npolshakova

assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,372 @@
...
+       provider:
+         openai:
+           model: mock-llm
+         host: mock-llm-predictor.kserve-test.svc.cluster.local
+         port: 80
+         path: "/llm/chat/completions"
+   EOF
+   ```
+
+2. Create a second `HTTPRoute` that routes to the `AgentgatewayBackend`. This route uses the same hostname as the KServe-generated route but matches only the `/llm/chat/completions` path, so the gateway prefers it for LLM traffic.
```
danehans: I'm still coming up to speed on kserve. Can we use the route generated by kserve, even if it's configured with a broader path match or tell the user to update the kserve-generated route by providing a `kubectl patch` command?

assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,372 @@
...
+               value: /llm/chat/completions
+         backendRefs:
+           - name: mock-llm-backend
+             namespace: kserve-test
+             group: agentgateway.dev
+             kind: AgentgatewayBackend
+   EOF
+   ```
+
+
```
danehans, line 352: nit:  only 1 newline is needed.

Add a section to test inference before applying policy.

assets/agw-docs/pages/integrations/kserve.md
```diff
@@ -0,0 +1,372 @@
...
+## Step 6: Apply token-based rate limiting
```
danehans: We should call out that this section is optional unless you update the doc with a title that is specific to token-based rate limiting.

Lasse4: changed in https://github.com/agentgateway/website/pull/443/changes/01ce6469b594335125287e14aed5ae7edfa29757

assets/agw-docs/pages/integrations/kserve.md
Lasse4: added in https://github.com/agentgateway/website/pull/443/changes/01ce6469b594335125287e14aed5ae7edfa29757

assets/agw-docs/pages/integrations/kserve.md
Lasse4: added in https://github.com/agentgateway/website/pull/443/changes/01ce6469b594335125287e14aed5ae7edfa29757 :)

assets/agw-docs/pages/integrations/kserve.md
Lasse4: changed in https://github.com/agentgateway/website/pull/443/changes/01ce6469b594335125287e14aed5ae7edfa29757

assets/agw-docs/pages/integrations/kserve.md
Lasse4: If you update the HTTPRoute using a patch command, the KServe controller will immediately overwrite it and revert it to the default state.

npolshakova (approved):
LGTM!

artberger (approved):

PR #473, opened 2026-05-07
agw-standalone feedback: security/backend-tls - clarify service ide...

artberger:
Fixed misleading documentation about backend TLS service identity behavior in standalone mode.

Closes [#2432](https://github.com/solo-io/docs/issues/2432)


artberger:
This Subject Alternative Name section should be removed from the standalone section of the agentgateway docs. Make sure that we cover it in the relevant Kubernetes section, such as `content/docs/kubernetes/{VERSION}/security/backendtls/`

original feedback that this issue is supposed to address: 
```
"In Kubernetes environments, Service SANs are automatically populated from the service identity" is wrong/misleading. Actually:

We default to service hostname (k8s service hostname or backend static hostname)
This is not about automatic mesh TLS behavior via policy
Service identity + SPIFFE example suggests something mesh-like that isn't true for backend TLS policy
```

kristin-kronstain-brown (approved):

PR #538, opened 2026-06-01
Fix Static MCP backend manifest when using backendRef Service

### LINKED ISSUE #537: Wrong MCP static service backend yaml manifest
Wrong MCP backend yaml manifest when configuring static with backendRef service.
Currently it is:
```
apiVersion: agentgateway.dev/v1alpha1
kind: AgentgatewayBackend
metadata:
  name: mcp-backend
spec:
  mcp:
    targets:
    - name: mcp-target
      backendRef:
        name: mcp-website-fetcher
      port: 80
      protocol: SSE
```
It should be:
```
apiVersion: agentgateway.dev/v1alpha1
kind: AgentgatewayBackend
metadata:
  name: mcp-backend
spec:
  mcp:
    targets:
    - name: mcp-target
      static:
        backendRef:
          name: mcp-website-fetcher
        port: 80
        protocol: SSE
```

AhmadMasry:
Fixes #537

### DIFF
--- a/assets/agw-docs/pages/agentgateway/mcp/static.md
+++ b/assets/agw-docs/pages/agentgateway/mcp/static.md
@@ -106,10 +106,11 @@ spec:
   mcp:
     targets:
     - name: mcp-target
-      backendRef:
-        name: mcp-website-fetcher
-      port: 80
-      protocol: SSE
+      static:
+        backendRef:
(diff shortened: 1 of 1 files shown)


artberger (approved):
Thanks for the fix!

PR #546, opened 2026-06-03
Fixed test failure for DFP

kristin-kronstain-brown:
This change fixes the recurring CI failure in the Dynamic Forward Proxy doc test, which was caused by the test depending on the live public httpbin.org. That host is a free, frequently-overloaded service, and it returned 503 Service Unavailable from its own load balancer, turning the test red for reasons unrelated to agentgateway or the docs. Because a CI test can only be reliable if it depends solely on resources under its control, the hidden test assertion now targets the in-cluster httpbin service (httpbin.httpbin.svc.cluster.local:8000) instead of the public internet, making it hermetic and immune to third-party outages. The visible documentation still demonstrates egress to an external host (httpbin.org) so readers understand the real use case for a forward proxy, and a callout points to the in-cluster address for anyone who wants to verify the proxy without relying on an external site. This change also corrects a pre-existing error in the page's example commands, which used port 8080 directly against the LoadBalancer (it listens on 80) and omitted the kubectl port-forward step in the local-testing tab, both of which caused connection-refused errors when followed by hand.

### DIFF
--- a/assets/agw-docs/pages/traffic-management/dfp.md
+++ b/assets/agw-docs/pages/traffic-management/dfp.md
@@ -92,7 +92,7 @@ YAMLTest -f - <<'EOF'
 EOF
 {{< /doc-test >}}
 
-3. Send a request to a hostname of your choice, such as `httpbin.org`. The Dynamic Forward Proxy resolves the host at request time and forwards the request to it, so the host's welcome page is returned. Because no upstream hosts are pre-defined in the Backend, you can send a request to any reachable host without changing the configuration.
+3. Send a request to a hostname of your choice, such as `httpbin.org`. The Dynamic Forward Proxy resolves the host at request time and forwards the request to it, so the host's welcome page is returned. Because no upstream hosts are pre-defined in the Backend, you can send a request to any reachable host without changing the configuration. For example, for quick testing, you might send a request with the host header set to an in-cluster service, such as httpbin.httpbin.svc.cluster.local:8000.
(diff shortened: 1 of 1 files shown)


assets/agw-docs/pages/traffic-management/dfp.md
```diff
@@ -78,33 +78,42 @@ EOF
...
+   2. Send a request to the gateway proxy.
+      ```sh
+      curl -vik http://localhost:8080 -H "host: httpbin.org"
+      ```
    {{% /tab %}}
    {{< /tabs >}}
+
+   {{< callout type="info" >}}
+   The Dynamic Forward Proxy resolves any reachable host, including in-cluster services. To verify the proxy without depending on an external site, send a request with the host header set to an in-cluster service, such as `httpbin.httpbin.svc.cluster.local:8000`.
+   {{< /callout >}}
```
artberger: Not sure we really need this note? or possibly could just add it briefly to the step above, something like:

> Send a request to a hostname of your choice, such as `httpbin.org`. The Dynamic Forward Proxy resolves the host at request time and forwards the request to it, so the host's welcome page is returned. Because no upstream hosts are pre-defined in the Backend, you can send a request to any reachable host without changing the configuration. For example, for quick testing, you might send a request with the host header set to an in-cluster service, such as `httpbin.httpbin.svc.cluster.local:8000`.

artberger (approved):

assets/agw-docs/pages/traffic-management/dfp.md
kristin-kronstain-brown: Removed note and updated step 3 text.

PR #562, opened 2026-06-04
Add config explorer for standalone

### LINKED ISSUE #167: Document Config Spec as YAML
The JSON schema is located here: https://github.com/agentgateway/agentgateway/blob/main/schema/local.json

But it would be great to document the full spec in YAML format, so it's easy to see how to configure in YAML. Maybe this would be a generated YAML showing all of the configuration options.

howardjohn:
fixes https://github.com/agentgateway/agentgateway/issues/167


artberger (approved):
Thanks for adding this!

PR #581, opened 2026-06-09
agw-standalone feedback: llm/transformations - same as prompt templ...

artberger:
Remove the misleading Hugo alias from both standalone transformations.md files (latest and main).

Closes #2451


tools.go
```diff
@@ -0,0 +1,3 @@
+package tools
```
artberger: do we need to change the tools.go file?

PR #642, opened 2026-06-18
Fix hugo permissions

howardjohn:
Hugo 0.161 runs PostCSS under Node filesystem permissions. Autoprefixer calls Browserslist, and Browserslist tries to walk parent directories looking for two optional files: Browserslist config and `browserslist-stats.json`. That walk reaches `..`, which Node is not allowed to read, so the build panics.

This change makes those implicit lookups explicit:

- `overrideBrowserslist` gives Autoprefixer the browser targets directly, so it does not need to search for config.
- `stats: {}` prevents the optional custom-usage stats lookup.
- The output intent is unchanged: production CSS still runs through Autoprefixer.


artberger (approved):
Thanks for the update. We might move this into the shared theme, but will merge it here for now.

PR #652, opened 2026-06-18
docs/minor fixes

howardjohn:
- **Minor UI fixes**
- **Remove 'up arrow' that is always masked by 'Ask AI'**


assets/css/custom.css
```diff
@@ -337,6 +337,10 @@ h6 {
     border-left-color: #374151 !important;
 }
 
+#solo-back-to-top {
+    display: none !important;
+}
+
```
artberger, line 343: I am not sure what this does? if it's trying to remove the `Scroll to top` in the right-hand TOC, I still see that locally (and not sure why we'd remove it?)

Image

howardjohn: it's this other scroll to top button that is behind the ask ai on every page so you can't click it anyways. it's blue

artberger (approved):

PR #646, opened 2026-06-18
Docs - add playwright for UI screenshots

artberger:
Use playwright to capture UI screenshots for various doc guides. 
- When possible, use the existing doc-tests framework to create the environment of agentgateway and related resources. 
- Support light and dark mode.
- Adds a workflow for running the refresh nightly.

Example PR that uses this to start capturing screenshots for the 1.3 release: https://github.com/agentgateway/website/pull/645

PR #672, opened 2026-06-25
Playwright enhancements for kube UI

artberger:
update playwright, followup to standalone in #646 

- handle versioning
- add workflow for k8s mode, not just standalone, reusing doc-tests environment


playwright/scripts/sync-docs-images.mjs
```diff
@@ -1,54 +1,115 @@
...
       continue;
     }
 
-    console.log(`${dryRun ? '[dry-run] ' : ''}${basename(src)} -> ${dest}`);
-    if (!dryRun) copyFileSync(src, resolve(repoRoot, dest));
-    copied++;
+    if (VERSION === 'main') {
+      // Publish to assets/img/main/<name> only when main visually differs from latest.
+      const latestBaseline = findBaseline(name, `latest-${variant}`);
+      if (!diverged(baseline, latestBaseline)) {
```
artberger, line 88: **[Major]** When DOC_VERSION=main and the new main baseline no longer diverges from the latest baseline, the script increments shared and continues without removing the existing assets/img/main/<name>.png target. Once any main-only image has been published, a later run where the images converge will leave that stale file committed, so any version split that points at img/main/ keeps showing an outdated screenshot even though the workflow reported it as shared with latest.
**Suggestion**: Before continuing in the non-diverged branch, compute mainDest(dest) and delete that target if it exists, so create-pull-request can commit the removal alongside the refreshed baselines.

artberger (commented):
Found one issue in the screenshot sync path: stale main-version images are not removed when they stop diverging from latest. (code quality 8.5/10, production).
_Automated review by the multi-agent swarm — initiated and approved by @artberger._

playwright/scripts/sync-docs-images.mjs
artberger: Fixed in 7b2b930. When `DOC_VERSION=main` and a baseline no longer diverges from latest, the non-diverged branch now computes `mainDest(dest)` and `rmSync`es it if it exists (logged as `CONVERGED rm ...`), so `create-pull-request` commits the removal alongside the refreshed baselines. Added a `convergedImages` summary that lists those images and reminds the author to drop any `{{< version >}}` split that pointed `main` at `img/main/`. Verified the script still parses and the `main` dry-run runs clean.

kristin-kronstain-brown (approved):

PR #673, opened 2026-06-25
docs: fix mcp authz examples

jgreeer:
Despite the MCP target policy docs stating that mcp authz is a per backend policy not per target, the mcp authz doc examples create per target policy. 
https://github.com/agentgateway/website/blob/c640918ca99a68701f9ff319056528553161fc10/content/docs/standalone/main/mcp/mcp-target-policies.md?plain=1#L34

This PR fixes the docs, also in local testing I found you have to add `audiences` to the mcp auth policy. We also rewrite the per target example to use CEL expressions.


artberger (approved):
Thanks for the fix!

PR #674, opened 2026-06-25
docs: update KServe install checks

danehans:
- Verify the KServe Gateway with `kubectl get gateway` and refresh the example output.
- Update the KServe Helm chart version to `v0.19.0`.
- Add a KServe controller availability check after the Helm install step.


artberger (approved):
Thanks for the update!

PR #732, opened 2026-07-13
docs: add Request Log page documenting SQLite and PostgreSQL database…

guenhter:
… backends


artberger (approved):
Thanks for the contribution!

PR #748, opened 2026-07-16
Add preview workflows

artberger:
- updates PRs to use workflows to build previews, instead of relying on Cloudflare
- cleanup workflow runs on PR close as well as daily to clean up any stale preview branches
- multi-workflow approach so that forks can get preview links, too

### DIFF
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -1,0 +1,88 @@
+# Preview build — untrusted half of the fork-preview pipeline.
+#
+# WHY TWO WORKFLOWS:
+# Cloudflare Pages' native Git integration does not build preview deployments
+# for PRs opened from forks: a fork cannot be trusted with the Cloudflare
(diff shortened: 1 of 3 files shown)


artberger:
Closing per some concerns, will just have previews as-is for contributors with write permissions on branches in clones of the repo vs. forks.

Concerns and possibly ways to address in the future:

If you ever do need true fork previews (e.g., contributor volume grows), the version worth building is the minimal-blast-radius one:
1. Never trust the artifact for the PR number — derive it from the trusted github.event.workflow_run.head_sha via an API lookup (or at minimum validate ^[0-9]+$ before any use). This kills the injection.
2. Separate Cloudflare Pages project for fork previews, no production custom domain attached — so attacker-controlled HTML can't sit next to production.
3. Denylist Pages control files (_worker.js, functions/, _headers, _redirects) — Hugo never emits them, so their presence is a tamper signal, and _worker.js/functions/ is the difference between "serving static HTML" and "running attacker code in your Pages project."

PR #750, opened 2026-07-16
blog: add semantic routing cost guide

danehans:
## Summary
- add a reproducible cost-based semantic-routing guide based on an example developer user case
- include the validated catalog-priced evaluation, flow diagram, and SVG report
- cover optional cache-transition evaluation and model-tier overrides

## Observed run
- 100 primary requests with exact agentgateway model-catalog lookups
- $0.914840 always-expensive versus $0.591444 routed: 35.4% lower spend
- 30 nano and 20 GPT-5.5 selections; 20 of 25 complex prompts escalated to GPT-5.5

## Validation
- `hugo --minify --destination <temporary-directory> --cleanDestinationDir` with the bundled Node runtime

PR #757, opened 2026-07-17
docs: update vLLM Semantic Router integration

danehans:
## Summary

- add a focused vLLM Semantic Router integration overview
- distinguish vSR from the vLLM inference provider integration
- link to the upstream deployment guide and agentgateway cost-routing example
- add a shared architecture diagram with explicit external-processing and backend-routing boundaries
- rename the page slug to `vllm-semantic-router` while preserving `vllm` as an alias

## Validation

- `git diff --check`
- `xmllint --noout assets/img/integrations/vllm-semantic-router-cost-routing.svg`
- verified both external resource links return HTTP 200
- `hugo160 --gc --minify`


danehans:
Converted to "draft" since it dep's on the referenced [blog post](https://github.com/agentgateway/website/pull/750).

artberger (approved):

PR #774, opened 2026-07-23
ConfigMap with hashed key

artberger:
Update virtual keys and API keys guides with ConfigMap and hashed key-based steps

PR #775, opened 2026-07-23
Guardrail updates

kristin-kronstain-brown:


content/docs/kubernetes/main/llm/guardrails/bedrock-guardrails.md
```diff
@@ -156,12 +156,56 @@ AWS Bedrock Guardrails are model-agnostic and can be applied to any Large Langua
...
+The `policies` field configures how agentgateway connects and authenticates to the AWS Bedrock Guardrails service when it evaluates a request or response.
+
+### Authentication
+
+Under `policies.auth`, set one credential source (`aws`, `secretRef`, or `key`). Optionally, set `location` to control where the credential is placed.
+
+| Method | Description |
+| -- | -- |
+| `aws` | Authenticate with AWS credentials. Set `aws: {}` to use the default AWS credential chain (IAM role, environment variables, or instance profile), or set `aws.secretRef` to read credentials from a Kubernetes secret. |
+| `secretRef` | Read the API key from a Kubernetes secret. By default, the key that matches the credential location is used, such as `Authorization` for the default header location. To use a different key, set `secretRef.key`. |
```
artberger, line 172: In my PR, I updated the preferred storage method for API keys from Secrets to hashed ConfigMaps...perhaps the example here could use CM instead? 

https://github.com/agentgateway/website/pull/774/

content/docs/kubernetes/main/llm/guardrails/bedrock-guardrails.md
```diff
@@ -156,12 +156,56 @@ AWS Bedrock Guardrails are model-agnostic and can be applied to any Large Langua
    The request was rejected due to inappropriate content
    ```
 
+## Backend connection and authentication policies
```
artberger, line 161: So there are policies already in the config examples earlier in this guide (and the others). Perhaps we crosslink to this section down here for more policy options, and summarize briefly why they would want to set these.

content/docs/kubernetes/main/llm/guardrails/bedrock-guardrails.md
kristin-kronstain-brown: Claude pushed back on this one:

> Art's hashed-ConfigMap pattern (PR #774) is for inbound API-key authentication — apiKeyAuthentication, where the gateway hashes the client's presented key and compares it to a stored keyHash. It never needs the plaintext key.
> 
> The guardrail policies.auth credentials are the opposite direction: they're outbound credentials the gateway must transmit to Bedrock / OpenAI (the AWS keys for SigV4, or a bearer token in the Authorization header). I confirmed in the merged code:
> 
> - keyHash appears only in the inbound path (crates/agentgateway/src/http/apikey.rs, traffic_plugin.go) — never in guardrail auth.
> - Guardrail secretRef is a LocalSecretKeyRef that reads the actual credential value to send.
>
> So a hashed ConfigMap can't work here — you can't send a hash as a credential — and a plaintext ConfigMap would be strictly less secure than a Secret (ConfigMaps aren't confidential). The Secret in these examples is the correct pattern.

content/docs/kubernetes/main/llm/guardrails/bedrock-guardrails.md
kristin-kronstain-brown: Added this basic paragraph to each, though each one subs out the thing being connected to. 

> The `policies` field supports more than the `aws` credential source shown here. You can choose a different authentication method or tune the connection that agentgateway opens to Bedrock, such as setting a request timeout or custom TLS. For all the options, see [Backend connection and authentication policies](#backend-connection-and-authentication-policies).

artberger (approved):

PR #745, opened 2026-07-16
Docs for cross app access in standalone

artberger:
Docs for https://github.com/agentgateway/agentgateway/pull/2436, examples based off https://github.com/agentgateway/agentgateway/pull/2449

PR #777, opened 2026-07-23
Add XAA to kube mode docs

artberger:
Follow up to #745 , but for kube mode

### DIFF
--- a/assets/agw-docs/pages/security/backend-authn-cross-app-access.md
+++ b/assets/agw-docs/pages/security/backend-authn-cross-app-access.md
@@ -24,6 +24,9 @@ flowchart LR
 
 Cross App Access differs from [OAuth token exchange]({{< link-hextra path="/security/backend-authn-oauth/" >}}) in that it crosses a trust boundary: the IdP and the resource's authorization server are separate parties, so the gateway performs two exchanges and holds two client registrations, one at each token endpoint. For a single-leg exchange at one authorization server, use `oauthTokenExchange` instead.
 
+> [!NOTE]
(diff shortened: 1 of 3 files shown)


assets/agw-docs/pages/security/backend-authn-cross-app-access.md
```diff
@@ -0,0 +1,428 @@
...
+flowchart LR
+    Client -- ID token --> AGW[Agentgateway]
+    AGW -- "(a) token exchange" --> IdP[Identity provider]
+    IdP -- ID-JAG --> AGW
+    AGW -- "(b) jwt-bearer grant" --> RAS[Resource authorization server]
+    RAS -- access token --> AGW
+    AGW -- "Authorization: Bearer<br>access token" --> API[Downstream API]
+```
+
+Cross App Access differs from [OAuth token exchange]({{< link-hextra path="/security/backend-authn-oauth/" >}}) in that it crosses a trust boundary: the IdP and the resource's authorization server are separate parties, so the gateway performs two exchanges and holds two client registrations, one at each token endpoint. For a single-leg exchange at one authorization server, use `oauthTokenExchange` instead.
```
kristin-kronstain-brown, line 25: I don't know if this is a great suggestion since the doc already calls it out, but thought maybe couldn't hurt to be extra explicit.

> The About section defines XAA by its trust boundary — "the IdP and the resource's authorization server are separate parties" — but the walkthrough uses a single Keycloak acting as both. The doc states the collapse but never ties it back to that concept or offers a real-topology path. The standalone doc resolves this with a "Try it against hosted providers" section (xaa.dev, okta-auth0); the kube doc drops it, so the reader is left with a demo that structurally contradicts the concept being taught.
> 
> Fix: Add a short callout (or a brief section) after the Keycloak deploy step that both acknowledges the simplification and points to the real examples. Something like:
> 
> [!NOTE]
> To keep the demo self-contained, this single Keycloak plays both parties — the user's IdP and the resource authorization server. In production these are separate trust domains (that separation is the whole point of Cross App Access). For examples against genuinely separate providers, see the [traffic-cross-app-access examples](https://github.com/agentgateway/agentgateway/tree/main/examples/traffic-cross-app-access): xaa-dev (hosted IdenX IdP + hosted resource AS) and okta-auth0 (Okta IdP + Auth0 resource AS). These illustrate the two-party topology; the configs are written for config/standalone mode.

kristin-kronstain-brown (approved):

PR #815, opened 2026-07-29
Docs: Add comprehensive guide for authorization policies

### LINKED ISSUE #341: Agentgateway on Kubernetes security authorization docs
We should get a page under https://agentgateway.dev/docs/kubernetes/latest/security/ for Authorization; we don't actually have any general docs for it just for LLM and MCP specifically.

In that, we can include https://github.com/agentgateway/agentgateway/pull/1223 (when it merges) and specifically cover that `Deny` is risky https://github.com/agentgateway/agentgateway/pull/1223/changes#diff-187a1437b6cb0175d9afb601568815f913dcbdb840f87149f6455cc31f17a622R9

Roniscend:
This PR introduces new documentation for configuring authorization policies in Agentgateway.
 
 Fixes : #341


artberger:
Thank you for opening this PR. Here are some global comments:

**Concept**:
I like the intro info that you have about how the authorization policy works. We might be able to add some crosslinks to examples later in the guide that illustrate these concepts. There's also a comment from the copilot review about `Deny` not being covered.

**Guide**:
Usually we like to have a short end-to-end guide that has a `Before you begin` section with any prereqs such as a shortcode reference to the [prereq snippet](https://github.com/agentgateway/website/blob/main/assets/agw-docs/snippets/prereq.md): `{{< reuse "agw-docs/snippets/prereq.md" >}}`. Then ordered steps that walk users through setting up the feature and testing it with the sample app. These steps would walk them through, for example, how they would get the JWT if the authz expects one and pass it in a request that is authorized and show success, or not authorized and show the expected error.

**Reorganize the current topics**:

So there are 2 examples, 1 "using deny" topic, and then 2 use cases.

What we do in other places is have an H2 called `## More examples`

and then we could put all of these under there, and not have the colons, like:

```
### Allow requests by JWT claim
...

### Require a condition across all rules
...

### Deny policies
(keep the warning but provide an example config)

### MCP authorization
....

### JWT authorization
....
```

Roniscend:
> Thank you for opening this PR. Here are some global comments:
> 
> **Concept**: I like the intro info that you have about how the authorization policy works. We might be able to add some crosslinks to examples later in the guide that illustrate these concepts. There's also a comment from the copilot review about `Deny` not being covered.
> 
> **Guide**: Usually we like to have a short end-to-end guide that has a `Before you begin` section with any prereqs such as a shortcode reference to the [prereq snippet](https://github.com/agentgateway/website/blob/main/assets/agw-docs/snippets/prereq.md): `{{< reuse "agw-docs/snippets/prereq.md" >}}`. Then ordered steps that walk users through setting up the feature and testing it with the sample app. These steps would walk them through, for example, how they would get the JWT if the authz expects one and pass it in a request that is authorized and show success, or not authorized and show the expected error.
> 
> **Reorganize the current topics**:
> 
> So there are 2 examples, 1 "using deny" topic, and then 2 use cases.
> 
> What we do in other places is have an H2 called `## More examples`
> 
> and then we could put all of these under there, and not have the colons, like:
> 
> ```
> ### Allow requests by JWT claim
> ...
> 
> ### Require a condition across all rules
> ...
> 
> ### Deny policies
> (keep the warning but provide an example config)
> 
> ### MCP authorization
> ....
> 
> ### JWT authorization
> ....
> ```

Thanks for the review I will work on this make the changes and update this pr

Roniscend:
@artberger I have made the changes can you re-review this pr 
Thanks

assets/agw-docs/pages/security/authorization.md
```diff
@@ -0,0 +1,220 @@
...
+  name: authz-guide
+  namespace: {{< reuse "agw-docs/snippets/namespace.md" >}}
+spec:
+  targetRefs:
+  - group: gateway.networking.k8s.io
+    kind: Gateway
+    name: agentgateway-proxy
+  traffic:
+    jwtAuthentication:
+      providers:
```
artberger, line 72: This yaml fails, are you sure that these are the fields? the API does not seem to have it: https://agentgateway.dev/docs/kubernetes/latest/reference/api/#jwtprovider

```
strict decoding error: unknown field "spec.traffic.jwtAuthentication.providers[0].jwksUri",
unknown field "spec.traffic.jwtAuthentication.providers[0].name"
```

assets/agw-docs/pages/security/authorization.md
```diff
@@ -0,0 +1,220 @@
...
+        - name: example-provider
+          issuer: "https://example.com"
+          jwksUri: "https://example.com/.well-known/jwks.json"
+    authorization:
+      action: Allow
+      policy:
+        matchExpressions:
+          - "jwt.claims['group'] == 'eng'"
```
artberger: Is this valid CEL? `jwt.claims['group']`  it silently denies everything. Claims are top-level on jwt. Same token, same policy, only the expression differs:

```
┌──────────────────────────────┬──────────────────────────┐
│          Expression          │ Result with matching JWT │
├──────────────────────────────┼──────────────────────────┤
│ jwt.claims['sub'] == 'alice' │ 403                      │
├──────────────────────────────┼──────────────────────────┤
│ jwt.sub == 'alice'           │ 200                      │
└──────────────────────────────┴──────────────────────────┘
```

also used later in 2 of 4 examples.

assets/agw-docs/pages/security/authorization.md
```diff
@@ -0,0 +1,220 @@
...
+  targetRefs:
+  - group: gateway.networking.k8s.io
+    kind: Gateway
+    name: agentgateway-proxy
+  traffic:
+    authorization:
+      action: Require
+      policy:
+        matchExpressions:
+          - "source.address in ['10.0.0.0/8']"
```
artberger: `source.address in ['10.0.0.0/8']` can you verify that this works?

When I tried, it didn't seem to evaluate as true. `source.address` is IP-typed, not a string (`source.address.startsWith('10.')` → 403; `source.address == source.address` → 200), so in against CIDR strings can never match:

- As Require (the PR's example) → denies 100% of traffic.
- As Deny (the PR's example) → blocks nothing, returns 200.

maybe something like `cidr('10.0.0.0/8').containsIP(source.address)` (as Require with 0.0.0.0/0 → 200; as Deny → 403).

assets/agw-docs/pages/security/authorization.md
```diff
@@ -0,0 +1,220 @@
...
+### 2. Test the policy
+
+Get the gateway URL:
+
+{{< reuse "agw-docs/snippets/agw-get-gateway-url-k8s.md" >}}
+
+Make a request without a JWT to see the expected error:
+
+```bash
+curl -i http://$INGRESS_GW_ADDRESS/
```
artberger: For the curls...

- we typically have tabs for Cloud Provider LoadBalancer vs local port-forward, see other guides for examples
- all the curls omit the host header that the sample app that this page uses in the prereqs requires. The prereq's sample-app route binds www.example.com, so all three requests return 404, not the documented 401/403/200. Verified: no host → 404, `-H "host: www.example.com"` → 200.

assets/agw-docs/pages/security/authorization.md
```diff
@@ -0,0 +1,220 @@
...
+```
+
+**Output:**
```
artberger: we typically phrase it as 

```
Example output:
```

assets/agw-docs/pages/security/authorization.md
```diff
@@ -0,0 +1,220 @@
...
+```
+HTTP/1.1 401 Unauthorized
```
artberger: When you go back and test each of these, make sure that each scenario yields the expected status code, some of these might need to change.

assets/agw-docs/pages/security/authorization.md
```diff
@@ -0,0 +1,220 @@
...
+```
+HTTP/1.1 200 OK
+...
+```
+
+## More examples
+
+### Allow requests by JWT claim
+
+The following example uses an `Allow` action to grant access to requests that include a specific JWT claim. Requests without a matching JWT or with an incorrect claim are denied.
```
artberger: The "Allow requests by JWT claim" example targets the Gateway, which locks it down. But it doesn't have the jwtAuthentication, so jwt is unpopulated, which means all traffic yields 403s, while the policy reports Valid Attached with no error surfaced. The dependency is explained later in the Jwt authorization section, but it might be confusing here to give an example that locks everything. Maybe combine the sections and link out to the guides for more.

assets/agw-docs/pages/security/authorization.md
```diff
@@ -0,0 +1,220 @@
...
+      policy:
+        matchExpressions:
+          - "jwt.claims['group'] == 'eng'"
+EOF
+```
+
+| Field | Description |
+|---|---|
+| `action` | The authorization action to take when the expressions match. Use `Allow` for allow-listing. |
+| `policy.matchExpressions` | A list of CEL expressions. All expressions must evaluate to `true` for the rule to match. |
```
artberger: Line 22 says "at least one expression"; the field table at line 146 says "All expressions must evaluate to true."

So I think L146 is wrong because Allow expressions are OR'd, Require AND'd.

or if you test it, I think this is the result you would get:
```
┌─────────┬──────────────┬───────────────┬─────────────┐
│         │ [true,false] │ [false,false] │ [true,true] │
├─────────┼──────────────┼───────────────┼─────────────┤
│ Allow   │ 200          │ 403           │ —           │
├─────────┼──────────────┼───────────────┼─────────────┤
│ Require │ 403          │ —             │ 200         │
└─────────┴──────────────┴───────────────┴─────────────┘
```

assets/agw-docs/pages/security/authorization.md
```diff
@@ -0,0 +1,115 @@
...
+| `Require` | All `Require` expressions must evaluate to `true` for the request to proceed. If any expression evaluates to `false`, the request is denied. Use `Require` to add mandatory conditions that must be satisfied alongside any `Allow` rules. |
+| `Deny` | Denies access when the expression matches. See the [warning below](#using-deny-policies). |
+
+### Evaluation order
+
+When an {{< reuse "agw-docs/snippets/policy.md" >}} is applied to a Gateway, it processes authorization as follows:
+
+1. All `Require` rules are evaluated first. If any `Require` expression is `false`, the request is denied immediately.
+2. `Allow` rules are evaluated next. At least one `Allow` expression must match.
+3. If no `Allow` rule matches, the request is denied.
```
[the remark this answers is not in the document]
artberger: also "If no Allow rule matches, the request is denied" only holds when at least one Allow rule is configured. A Require-only policy with all Requires true → 200. Make sure the eval order is reflected later in the example so they don't contradict

assets/agw-docs/pages/security/authorization.md
```diff
@@ -0,0 +1,220 @@
+Authorization policies in agentgateway let you control which requests are allowed to reach your backends. Policies apply across all traffic types — HTTP routes, LLM providers, MCP servers, and agents — giving you a unified way to enforce access rules.
+
+For a general overview of how policies are structured, see [Policy sections]({{< link-hextra path="/about/policies/overview/" >}}).
+
+## How authorization works
+
+Agentgateway uses the `authorization` field inside an {{< reuse "agw-docs/snippets/policy.md" >}} to evaluate whether an incoming request should be allowed or rejected. Authorization rules are expressed as [Common Expression Language (CEL)]({{< link-hextra path="/reference/cel/" >}}) expressions, which let you match on request headers, JWT claims, source IP addresses, MCP tool names, and more.
+
+The `authorization` field can appear in two places in a policy:
```
artberger: Can you verify that this policy cannot also belong in a 3rd place:

spec.frontend.networkAuthorization — L4 network authorization, which has the CIDR example for source address that you mention later.

The API docs mention that `Authorization` appears in 4 places:

https://agentgateway.dev/docs/kubernetes/latest/reference/api/#authorization

assets/agw-docs/pages/security/authorization.md
```diff
@@ -0,0 +1,220 @@
...
+```bash
+curl -i http://$INGRESS_GW_ADDRESS/ -H "Authorization: Bearer <VALID_JWT>"
+```
+
+**Output:**
+```
+HTTP/1.1 200 OK
+...
+```
+
```
artberger, line 231: We typically have a `## Cleanup` section, see other pages for examples.

content/docs/kubernetes/latest/security/authorization.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Authorization
+weight: 12
```
artberger: We typically do intervals of 5 for weighting, or if pages should have equal weight and just be sorted alphabetically, then the same weights. Maybe 15 would work.

artberger (commented):
Hi @Roniscend thanks for continuing to work on this PR. I have left some comments. Most are about making sure that the examples and steps actually work when you try them out, at least for me, I was running into some issues. It might help to follow the prereqs to set up a kind cluster, install agentgateway, and test the steps in order. 

We also have a test framework as described here: https://github.com/agentgateway/website/blob/main/scripts/TEST_FRAMEWORK.md

to write tests within the guides that run nightly to catch errors. Usually after you have the steps in good order, the tests are easy to write.

Roniscend:
Hey @artberger I have updated the pr , could you review it again 
Thanks

artberger (approved):

artberger:
Thank you! I will add the tests and merge.

PR #837, opened 2026-08-01
docs: clarify MCP guardrail mutation trust boundary

Poke-nushi:
## Summary

- Document the trust boundary for request-phase MCP guardrail processors that mutate JSON-RPC `params`.
- Clarify that request mutation does not change the selected backend and that CEL-based RBAC uses the originally resolved backend and identifier.
- Explain where authorization must run when a processor chain can mutate the effective request.

## Why

Request-phase processors can determine the effective request sent upstream. The documentation now makes the resulting operator responsibility explicit, including the ordering required when authorization must cover post-mutation identifiers or parameters.

## Validation

- `git diff --check`
- Static checks for the new trust-boundary section and authorization-ordering guidance
- Read-only review against the current MCP session, guardrail-chain, RBAC, and ExtMCP protocol implementation

The Hugo site build and framework tests were not run locally because Hugo is not installed in this environment.


artberger (approved):
Thanks for the contribution!

PR #845, opened 2026-08-03
docs: document configurable webhook guardrail headers and path (CEL)

robin-fiddler:
## Summary

- Documents the `webhook.headers` field added in [agentgateway/agentgateway#2595](https://github.com/agentgateway/agentgateway/pull/2595) (and exposed through the Kubernetes CRD/xDS API in [#2723](https://github.com/agentgateway/agentgateway/pull/2723)), which lets a prompt guard webhook set CEL-computed headers on the outgoing request, and override the default `/request` / `/response` path via the `:path` pseudo-header.
- Updates two guides that only showed the old bare `target`/`backendRef` webhook config with no mention of `headers`:
  - `content/docs/standalone/{main,latest}/llm/prompt-guards/webhooks.md` — standalone binary "Custom webhooks" guide.
  - `assets/agw-docs/pages/agentgateway/guardrails.md` — shared Kubernetes guardrail webhook guide (rendered under both `kubernetes/main` and `kubernetes/latest`).
- Each addition follows the existing `addRequestHeaders`/`transformation` CEL documentation style already used elsewhere (e.g. `tailscale.md`), and links to the `/reference/cel/` page.

## Validation

- Built the site locally with Hugo (`hugo --gc`) and verified both changed pages render correctly: new TOC entries appear, the `link-hextra` CEL reference link resolves to the correct versioned path, and the shared `guardrails.md` page renders identically under both `kubernetes/main` and `kubernetes/latest`.
- No content changes needed for `kubernetes/1.0.x`–`1.3.x` or `2.2.x`, since the CRD field only shipped in v1.4.1 (after those doc snapshots were cut) — only `main`/`latest` are documented, per the `hugo.yaml` version dropdown config.


artberger (approved):
Thanks for adding the docs for this feature!

PR #854, opened 2026-08-04
Clarify body buffer limit in ExtProc documentation

keithmattix:
Updated the body buffer limit description for request processing options.


artberger (approved):
Thank you!

PR #856, opened 2026-08-04
Misc fixes

artberger:
- rename shortcodes to get rid of the old `kgateway` name
- followup on some community PRs

### DIFF
--- a/assets/agw-docs/pages/install/advanced.md
+++ b/assets/agw-docs/pages/install/advanced.md
@@ -6,12 +6,12 @@ You can update several installation settings in your Helm values file. For examp
   helm show values {{< reuse "/agw-docs/snippets/helm-path.md" >}} --version {{< reuse "agw-docs/versions/helm-version-upgrade.md" >}}
   ```
 
-* **Get a file with all values**: You can get a `{{< reuse "/agw-docs/snippets/helm-kgateway.md" >}}/values.yaml` file for the upgrade version by pulling and inspecting the Helm chart locally.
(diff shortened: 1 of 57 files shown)


assets/agw-docs/pages/install/advanced.md
```diff
@@ -54,81 +54,109 @@ controller:
...
+     - key: kubernetes.io/metadata.name
+       operator: In
+       values:
+       - agentgateway-tenant-b-system
+   - matchLabels:
+       gateway-controller: tenant-b
+   ```
 
-Install the second control plane. Install the {{< reuse "agw-docs/snippets/kgateway.md" >}} and Kubernetes Gateway API custom resource definitions (CRDs) only once per cluster.
+   Each entry in `discoveryNamespaceSelectors` is disjunctive, so this example selects the control plane's own namespace and every namespace with the `gateway-controller: tenant-b` label. Include the control plane's own namespace, because the controller watches resources in that namespace, including the certificate that secures its xDS connection.
```
Rachael-Graham: maybe move the second sentence up to 83 so it doesnt get lost after the example explanation

assets/agw-docs/pages/observability/control-plane-metrics.md
```diff
@@ -4,39 +4,61 @@ By default, the {{< reuse "/agw-docs/snippets/kgateway.md" >}} control plane exp
...
 
-```yaml
-monitoring:
-  enabled: true
-  serviceMonitor:
-    enabled: true
-```YAML
-This creates:
+> [!IMPORTANT]
+> Install the Prometheus Operator custom resource definitions (CRDs) in your cluster before you enable monitoring. Without the CRDs, Prometheus cannot recognize the ServiceMonitor and PodMonitor resources that the chart creates.
```
Rachael-Graham: is there a link or command for installing the prom crds?

Rachael-Graham (approved):

PR #869, opened 2026-08-07
docs: cover frontend port/listener precedence, tie-breaks, and strategy.inheritance

### LINKED ISSUE #824: docs: policy merging page omits frontend port precedence, the same-specificity tie-break, and spec.strategy
**Page:** [Targeting and merging](https://agentgateway.dev/docs/kubernetes/main/about/policies/target-merge/) — source: `content/docs/kubernetes/main/about/policies/target-merge.md`

While writing an e2e test that adds a `llm.toolCalls` CEL attribute through `frontend.accessLog`, I hit behavior this page doesn't cover, and spent a debugging cycle because the page reads as if it does. The field-level/atomic merge model described here matches the implementation — the gaps below are all about **`frontend` precedence** and **what happens at equal specificity**. All source references are at tag `v1.4.1` (`163ea21`).

## 1. The `frontend` precedence row omits the port level (and the targeting table omits `port` entirely)

Today:

| Section | Precedence order (lowest to highest) |
| -- | -- |
| `frontend` | Field-level merge across policies that target the same Gateway. |

and the targeting table lists `frontend` as `Gateway` / sectionName `Not allowed` / "Applies to all listeners on the targeted Gateway."

But `AgentgatewayPolicy.spec.targetRefs[].port` exists and its own CRD description says it is **frontend-only**:

> The port of the target resource this policy applies to. At most one of `sectionName` or `port` may be set. Only valid on frontend policies targeting a `Gateway`.

So a `frontend` policy can be scoped narrower than "all listeners on the Gateway", and there is a real precedence order between the two. In `crates/agentgateway/src/store/binds.rs`, `frontend_policies()` resolves a port-scoped target by visiting port-targeted policies first and then chaining gateway-level ones to fill remaining fields, which the unit tests confirm:

- `frontend_policy_port_precedence` — a port-level policy wins over a gateway-level one.
- `frontend_policy_gateway_port_inherits_gateway_level` — with no port-level policy, gateway-level fields still apply to the port-scoped resolution.
- `frontend_policy_listener_precedence` — there is also a listener level, which wins over port.

Suggested fix: make the `frontend` row show the hierarchy (something like `Gateway < port < listener`) and mention `port` in the targeting table, since a reader who only sees "applies to all listeners" won't know port-scoping exists or how it interacts.

(Related question rather than a claim: the targeting table lists `Gateway` as the only valid target for `frontend`, but the code has a listener level reached via ListenerSet — `listenerset_targeted_policy_is_found_via_listener_frontend_policies`. If ListenerSet-targeted `frontend` policies are user-facing, the valid-targets column may need `ListenerSet` too.)

## 2. Nothing documents what happens when two policies of *equal* specificity set the same field

The only tie-break the page states is "If two policies set the same field, the more specific policy takes precedence." For `frontend` that rule frequently has no answer: two policies that both target the whole Gateway have identical specificity, and `frontend` has no `sectionName` to differentiate them.

What actually happens (observed on OSS v1.4.0 in kind, and consistent with the source): two `AgentgatewayPolicy` resources that each set `frontend.accessLog.attributes.add` on the same Gateway do **not** both contribute — one policy's entire `accessLog` section applies and the other's is dropped. Both policies report `Accepted=True` / `Attached=True` with no warning and no condition, so the only symptom is that the attributes silently never appear in the access log. Deleting the winning policy makes the other one's attributes appear immediately, confirming the loser was valid all along.

The selection is also not stable: `policies_by_target` is a `hashbrown::HashMap<PolicyTarget, HashSet<PolicyKey>>`, and `FrontendPolices::set_if_empty` uses `get_or_insert_with` per section, so the winner is whichever `PolicyKey` the `HashSet` happens to yield first — not creation time, name, or namespace. (agentgateway/agentgateway#1866 reports the same ordering property for the `backend.ai.*` path, where a maintainer confirmed the non-merging itself is currently intended.)

Suggested fix: say explicitly on this page that when two policies of the same specificity set the same field, one is chosen arbitrarily and the other is silently dropped with no status condition — and that users should therefore keep a given field/section in a single policy per target. That one sentence would have saved the debugging cycle, and it's useful guidance regardless of whether the underlying merge semantics change later.

## 3. `spec.strategy.inheritance` isn't mentioned on the merging page

`AgentgatewayPolicy.spec.strategy.inheritance` (`Default` | `Override`) directly controls merge/override semantics — per the CRD, `Override` "blocks traffic policies at more-specific attachment points from being included in the effective policy", it is valid only on `traffic` policies (CRD CEL rule: "strategy.inheritance may only be set on traffic policies"), and "Frontend and backend policy merging does not use inheritance."

That last clause is exactly the kind of thing readers come to this page for, but `strategy` appears nowhere in the repo's docs (searching `agentgateway/website` for `strategy.inheritance` returns no hits). It seems worth a short section here, if only to state that it exists, that it is traffic-only, and that it does not apply to `frontend`/`backend`.

kristin-kronstain-brown:
## Summary

Fixes gaps in the [Targeting and merging](https://agentgateway.dev/docs/kubernetes/main/about/policies/target-merge/) page reported in #824:

- The `frontend` targeting/precedence rows omitted `targetRefs[].port`, which scopes a frontend policy to a single listener port. Added `port` to the table and the precedence order (`Gateway < Port < Listener`), plus a new "Frontend section restrictions" table showing which frontend fields (`accessLog`, `metrics`, `tracing`) can also use `sectionName` to scope to a listener, and which can't.
- Nothing documented what happens when two equally-specific policies set the same field. Added a section explaining that agentgateway picks one policy's value arbitrarily and silently drops the rest, with no status condition indicating the drop.
- `spec.strategy.inheritance` wasn't mentioned anywhere on the page (or the repo). Added a section covering `Default` vs `Override`, and that it's valid only on `traffic` policies.

Verified all three behaviors against `agentgateway/agentgateway` at tag `v1.4.1` (source, unit tests, and the `AgentgatewayPolicy` CRD's CEL validation rules) before writing.

On the "related question" in the issue about whether `ListenerSet` should be a valid frontend target: the CRD's CEL validation (`the 'frontend' field can only target a Gateway`) confirms it isn't today, so the targeting table's valid-targets column is unchanged.

## Test plan

- [x] Built the full site with `hugo160` — no errors
- [x] Confirmed the new anchors (`#frontend-restrictions`, `#ties`, `#strategy`) and tables render correctly in the built HTML
- [x] Page front matter is `test: skip`, so no doc-test annotations apply

🤖 Generated with [Claude Code](https://claude.com/claude-code)


content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -18,10 +18,21 @@ The following table shows which resource types each policy section can target.
 
 | Policy section | Valid targets | sectionName | Notes |
 | -- | -- | -- | -- |
-| `frontend` | Gateway | Not allowed | Applies to all listeners on the targeted Gateway. |
+| `frontend` | Gateway | Depends on the field; see [Frontend section restrictions](#frontend-restrictions). | Applies to all listeners on the targeted Gateway. A `targetRef` or `targetSelector` can also set `port` to scope the policy to a single listener port on the Gateway. |
```
kristin-kronstain-brown:
```suggestion
| `frontend` | Gateway | Depends on the field. See [Frontend section restrictions](#frontend-restrictions) for more information. | Applies to all listeners on the targeted Gateway. A `targetRef` or `targetSelector` can also set `port` to scope the policy to a single listener port on the Gateway. |
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -18,10 +18,21 @@ The following table shows which resource types each policy section can target.
...
 | `traffic` | Gateway, HTTPRoute, GRPCRoute, ListenerSet | Optional | When targeting a Gateway, the `sectionName` selects a specific listener. When targeting an HTTPRoute or GRPCRoute, the `sectionName` selects a specific route rule. |
 | `backend` | Gateway, HTTPRoute, GRPCRoute, ListenerSet, Service, {{< reuse "agw-docs/snippets/backend.md" >}} | Optional | When targeting a Service, the `sectionName` selects a specific port. When targeting an {{< reuse "agw-docs/snippets/backend.md" >}}, the `sectionName` selects a specific sub-backend. |
 
+### Frontend section restrictions {#frontend-restrictions}
+
+`port` (to scope a `frontend` policy to a single listener port on the Gateway) is allowed for every `frontend` field. Whether a `frontend` policy can also set `sectionName` (to scope the policy to a single named listener) depends on which fields the policy sets.
```
kristin-kronstain-brown:
```suggestion
Use `port` to scope a `frontend` policy to a single listener port on the Gateway, which is allowed for every `frontend` field. Whether a `frontend` policy can also set `sectionName` to scope the policy to a single named listener depends on which fields the policy sets.
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -18,10 +18,21 @@ The following table shows which resource types each policy section can target.
...
+### Frontend section restrictions {#frontend-restrictions}
+
+Use `port` to scope a `frontend` policy to a single listener port on the Gateway, which is allowed for every `frontend` field. Whether a `frontend` policy can also set `sectionName` to scope the policy to a single named listener depends on which fields the policy sets.
+
+| Fields | `sectionName` |
+| -- | -- |
+| `accessLog`, `metrics`, `tracing` | Allowed. Selects a single listener. |
+| `connect`, `http`, `networkAuthorization`, `proxyProtocol`, `tcp`, `tls` | Not allowed. |
+
+If a policy sets any field from the second row, the Kubernetes API server rejects `sectionName` on that policy, even if the policy also sets a field from the first row.
```
kristin-kronstain-brown:
```suggestion
If a policy sets any not allowed field, the Kubernetes API server rejects `sectionName` on that policy, even if the policy also sets an allowed field.
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,31 @@ This field-level merge applies to all fields, including nested sub-fields. Each
 
 Conditional policies are selected first based on their conditions. Only the selected policies participate in merge precedence evaluation.
 
-Each policy section follows a different precedence order based on the specificity of the target. The more specific the target, the higher the precedence.
+Each policy section follows a different precedence order based on the specificity of the target. The more specific the target, the higher the precedence. In the following table, `A < B` means a policy attached at `B` overrides a matching field set by a policy attached at `A`.
```
kristin-kronstain-brown:
```suggestion
Each policy section follows a different precedence order based on the specificity of the target. The more specific the target, the higher the precedence. In the following table, `<` shows which policies override others.For example, `A < B` means a policy attached at `B` overrides a matching field set by a policy attached at `A`.
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,31 @@ This field-level merge applies to all fields, including nested sub-fields. Each
 
 Conditional policies are selected first based on their conditions. Only the selected policies participate in merge precedence evaluation.
 
-Each policy section follows a different precedence order based on the specificity of the target. The more specific the target, the higher the precedence.
+Each policy section follows a different precedence order based on the specificity of the target. The more specific the target, the higher the precedence. In the following table, `<` shows which policies override others.For example, `A < B` means a policy attached at `B` overrides a matching field set by a policy attached at `A`.
```
kristin-kronstain-brown:
```suggestion
Each policy section follows a different precedence order based on the specificity of the target. The more specific the target, the higher the precedence. In the following table, `<` shows which policies override others. For example, `A < B` means a policy attached at `B` overrides a matching field set by a policy attached at `A`.
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,31 @@ This field-level merge applies to all fields, including nested sub-fields. Each
...
+
+For example, if a Gateway-level policy sets `backend.tcp` and `backend.tls`, and a Backend-level policy sets `backend.tls`, the effective policy uses `tcp` from the Gateway policy and `tls` from the Backend policy.
+
+### When two policies have equal specificity {#ties}
+
+If multiple policies with the same specificity set the same field, agentgateway picks one policy's value for that field and silently drops the rest. The selection isn't based on creation time, name, or namespace, so which policy wins isn't predictable and can change between controller restarts. Every affected policy still reports `Accepted` and `Attached` status conditions as `True`, with no condition indicating that a field was dropped.
+
+This applies to every policy section, but it's easiest to hit with `frontend`. A `frontend` policy has at most three specificity levels: the Gateway, an optional `port`, and an optional listener `sectionName` for the fields where `sectionName` is allowed. Two policies that both target only the Gateway have no way to differentiate their specificity.
```
kristin-kronstain-brown:
```suggestion
This configuration applies to every policy section, but noticeable with `frontend`. A `frontend` policy has at most three specificity levels: the Gateway, an optional `port`, and an optional listener `sectionName` for the fields where `sectionName` is allowed. Two policies that both target only the Gateway have no way to differentiate their specificity.
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,31 @@ This field-level merge applies to all fields, including nested sub-fields. Each
...
+
+For example, if a Gateway-level policy sets `backend.tcp` and `backend.tls`, and a Backend-level policy sets `backend.tls`, the effective policy uses `tcp` from the Gateway policy and `tls` from the Backend policy.
+
+### When two policies have equal specificity {#ties}
+
+If multiple policies with the same specificity set the same field, agentgateway picks one policy's value for that field and silently drops the rest. The selection isn't based on creation time, name, or namespace, so which policy wins isn't predictable and can change between controller restarts. Every affected policy still reports `Accepted` and `Attached` status conditions as `True`, with no condition indicating that a field was dropped.
+
+This configuration applies to every policy section, but noticeable with `frontend`. A `frontend` policy has at most three specificity levels: the Gateway, an optional `port`, and an optional listener `sectionName` for the fields where `sectionName` is allowed. Two policies that both target only the Gateway have no way to differentiate their specificity.
+
+To avoid this, set a given field or section in only one policy per target.
```
kristin-kronstain-brown:
```suggestion
To avoid this configuration, set a given field or section in only one policy per target.
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,31 @@ This field-level merge applies to all fields, including nested sub-fields. Each
...
+
+To avoid this configuration, set a given field or section in only one policy per target.
+
+## Merge strategy overrides {#strategy}
+
+`AgentgatewayPolicy.spec.strategy.inheritance` changes how `traffic` policies merge. It's valid only on policies that set `traffic`. The Kubernetes API server rejects a policy that sets `inheritance` alongside `frontend` or `backend`. Frontend and backend policy merging always follows the specificity order described above and doesn't use `inheritance`.
+
+| Value | Behavior |
+| -- | -- |
+| `Default` | The default value. Fields from more-specific attachment points, such as routes and route rules, can override fields from less-specific attachment points, such as gateways and listeners. Use this to set a `traffic` default at the Gateway that specific routes can override. |
```
kristin-kronstain-brown:
```suggestion
| `Default` | The default value. Fields from more-specific attachment points, such as routes and route rules, can override fields from less-specific attachment points, such as gateways and listeners. Use this value to set a `traffic` default at the Gateway that specific routes can override. |
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,31 @@ This field-level merge applies to all fields, including nested sub-fields. Each
...
+To avoid this configuration, set a given field or section in only one policy per target.
+
+## Merge strategy overrides {#strategy}
+
+`AgentgatewayPolicy.spec.strategy.inheritance` changes how `traffic` policies merge. It's valid only on policies that set `traffic`. The Kubernetes API server rejects a policy that sets `inheritance` alongside `frontend` or `backend`. Frontend and backend policy merging always follows the specificity order described above and doesn't use `inheritance`.
+
+| Value | Behavior |
+| -- | -- |
+| `Default` | The default value. Fields from more-specific attachment points, such as routes and route rules, can override fields from less-specific attachment points, such as gateways and listeners. Use this value to set a `traffic` default at the Gateway that specific routes can override. |
+| `Override` | Blocks `traffic` policies at more-specific attachment points from contributing to the effective policy. Use this when a less-specific policy, such as one at the Gateway level, must stay authoritative for everything below it. |
```
kristin-kronstain-brown:
```suggestion
| `Override` | Blocks `traffic` policies at more-specific attachment points from contributing to the effective policy. Use this value when a less-specific policy, such as one at the Gateway level, must stay authoritative for everything below it. |
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -18,10 +18,21 @@ The following table shows which resource types each policy section can target.
...
+| `frontend` | Gateway | Depends on the field. See [Frontend section restrictions](#frontend-restrictions) for more information. | Applies to all listeners on the targeted Gateway. A `targetRef` or `targetSelector` can also set `port` to scope the policy to a single listener port on the Gateway. |
 | `traffic` | Gateway, HTTPRoute, GRPCRoute, ListenerSet | Optional | When targeting a Gateway, the `sectionName` selects a specific listener. When targeting an HTTPRoute or GRPCRoute, the `sectionName` selects a specific route rule. |
 | `backend` | Gateway, HTTPRoute, GRPCRoute, ListenerSet, Service, {{< reuse "agw-docs/snippets/backend.md" >}} | Optional | When targeting a Service, the `sectionName` selects a specific port. When targeting an {{< reuse "agw-docs/snippets/backend.md" >}}, the `sectionName` selects a specific sub-backend. |
 
+### Frontend section restrictions {#frontend-restrictions}
+
+Use `port` to scope a `frontend` policy to a single listener port on the Gateway, which is allowed for every `frontend` field. Whether a `frontend` policy can also set `sectionName` to scope the policy to a single named listener depends on which fields the policy sets.
```
Nadine2016:
```suggestion
Use the `port` field to scope a `frontend` policy to a single listener port on the Gateway. Whether a `frontend` policy can also set a `sectionName` to scope the policy to a single named listener depends on which fields the policy sets.
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -18,10 +18,21 @@ The following table shows which resource types each policy section can target.
...
+
+| Fields | `sectionName` |
+| -- | -- |
+| `accessLog`, `metrics`, `tracing` | Allowed. Selects a single listener. |
+| `connect`, `http`, `networkAuthorization`, `proxyProtocol`, `tcp`, `tls` | Not allowed. |
+
+If a policy sets any not allowed field, the Kubernetes API server rejects `sectionName` on that policy, even if the policy also sets an allowed field.
```
Nadine2016:
```suggestion
If a policy sets any not allowed field, the Kubernetes API server rejects the `sectionName` on that policy.
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,31 @@ This field-level merge applies to all fields, including nested sub-fields. Each
...
 | `traffic` | Gateway < Listener < Route < Route rule |
 | `backend` | Gateway < Listener < Route (targetRef) < Route rule (targetRef) < Backend (targetRef) < Backend (inline on the backend object) < Route backend ref (inline on the route) |
 
-For example, if a Gateway-level policy sets `backend.tcp` and `backend.tls`, and a Backend-level policy sets `backend.tls`, the effective policy uses `tcp` from the Gateway policy and `tls` from the Backend policy.
\ No newline at end of file
+For `frontend`, the Listener level applies only to `sectionName`. Only some `frontend` fields allow `sectionName` at all. See [Frontend section restrictions](#frontend-restrictions) for which ones.
+
+For example, if a Gateway-level policy sets `backend.tcp` and `backend.tls`, and a Backend-level policy sets `backend.tls`, the effective policy uses `tcp` from the Gateway policy and `tls` from the Backend policy.
+
+### When two policies have equal specificity {#ties}
```
Nadine2016:
```suggestion
### Equal specificity {#ties}
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,31 @@ This field-level merge applies to all fields, including nested sub-fields. Each
...
+
+This configuration applies to every policy section, but noticeable with `frontend`. A `frontend` policy has at most three specificity levels: the Gateway, an optional `port`, and an optional listener `sectionName` for the fields where `sectionName` is allowed. Two policies that both target only the Gateway have no way to differentiate their specificity.
+
+To avoid this configuration, set a given field or section in only one policy per target.
+
+## Merge strategy overrides {#strategy}
+
+`AgentgatewayPolicy.spec.strategy.inheritance` changes how `traffic` policies merge. It's valid only on policies that set `traffic`. The Kubernetes API server rejects a policy that sets `inheritance` alongside `frontend` or `backend`. Frontend and backend policy merging always follows the specificity order described above and doesn't use `inheritance`.
```
Nadine2016:
```suggestion
`AgentgatewayPolicy.spec.strategy.inheritance` changes how `traffic` policies merge. It's valid only on policies that set `traffic`. The Kubernetes API server rejects a policy that sets the `inheritance` block alongside the `frontend` or `backend` block. Frontend and backend policy merging always follows the specificity order described above and doesn't use `inheritance`.
```

Nadine2016: described above is a little vague. can we make it clearer what applies?

kristin-kronstain-brown: Changed to: 
> `AgentgatewayPolicy.spec.strategy.inheritance` changes how `traffic` policies merge. It's valid only on policies that set `traffic`. The Kubernetes API server rejects a policy that sets the `inheritance` block alongside the `frontend` or `backend` block. Frontend and backend policy merging always follows the `frontend` and `backend` precedence orders in the [Merge precedence](#merging-precedence) table and never uses `inheritance`.

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,31 @@ This field-level merge applies to all fields, including nested sub-fields. Each
...
+
+For example, if a Gateway-level policy sets `backend.tcp` and `backend.tls`, and a Backend-level policy sets `backend.tls`, the effective policy uses `tcp` from the Gateway policy and `tls` from the Backend policy.
+
+### When two policies have equal specificity {#ties}
+
+If multiple policies with the same specificity set the same field, agentgateway picks one policy's value for that field and silently drops the rest. The selection isn't based on creation time, name, or namespace, so which policy wins isn't predictable and can change between controller restarts. Every affected policy still reports `Accepted` and `Attached` status conditions as `True`, with no condition indicating that a field was dropped.
+
+This configuration applies to every policy section, but noticeable with `frontend`. A `frontend` policy has at most three specificity levels: the Gateway, an optional `port`, and an optional listener `sectionName` for the fields where `sectionName` is allowed. Two policies that both target only the Gateway have no way to differentiate their specificity.
```
Nadine2016: I am not sure I understand if I can't have two frontend policies targeting the same gateway, even if one targets a port and the other a sectionName for example. of if gateway, port, sectionName do not really matter here and I can't simply attach two frontend policies to the same gateway

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,31 @@ This field-level merge applies to all fields, including nested sub-fields. Each
...
+To avoid this configuration, set a given field or section in only one policy per target.
+
+## Merge strategy overrides {#strategy}
+
+`AgentgatewayPolicy.spec.strategy.inheritance` changes how `traffic` policies merge. It's valid only on policies that set `traffic`. The Kubernetes API server rejects a policy that sets `inheritance` alongside `frontend` or `backend`. Frontend and backend policy merging always follows the specificity order described above and doesn't use `inheritance`.
+
+| Value | Behavior |
+| -- | -- |
+| `Default` | The default value. Fields from more-specific attachment points, such as routes and route rules, can override fields from less-specific attachment points, such as gateways and listeners. Use this value to set a `traffic` default at the Gateway that specific routes can override. |
+| `Override` | Blocks `traffic` policies at more-specific attachment points from contributing to the effective policy. Use this value when a less-specific policy, such as one at the Gateway level, must stay authoritative for everything below it. |
```
Nadine2016: just from reading the description, you would think that override allows you to override, but it is actually not letting you overwrite. just want to make sure that this is correct

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,31 @@ This field-level merge applies to all fields, including nested sub-fields. Each
...
 | Section | Precedence order (lowest to highest) |
 | -- | -- |
-| `frontend` | Field-level merge across policies that target the same Gateway. |
+| `frontend` | Gateway < Port < Listener |
 | `traffic` | Gateway < Listener < Route < Route rule |
 | `backend` | Gateway < Listener < Route (targetRef) < Route rule (targetRef) < Backend (targetRef) < Backend (inline on the backend object) < Route backend ref (inline on the route) |
 
-For example, if a Gateway-level policy sets `backend.tcp` and `backend.tls`, and a Backend-level policy sets `backend.tls`, the effective policy uses `tcp` from the Gateway policy and `tls` from the Backend policy.
\ No newline at end of file
+For `frontend`, the Listener level applies only to `sectionName`. Only some `frontend` fields allow `sectionName` at all. See [Frontend section restrictions](#frontend-restrictions) for which ones.
```
Nadine2016:
```suggestion
For a `frontend` policy, you can only apply the policy at the Listener level when the `frontend` policy field supports setting a `sectionName`. For more information about the fields that support the `sectionName` setting, see [Frontend section restrictions](#frontend-restrictions).
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,31 @@ This field-level merge applies to all fields, including nested sub-fields. Each
...
+
+For example, if a Gateway-level policy sets `backend.tcp` and `backend.tls`, and a Backend-level policy sets `backend.tls`, the effective policy uses `tcp` from the Gateway policy and `tls` from the Backend policy.
```
Nadine2016: this example is a little weird here because it does not relate to the frontend policy above

content/docs/kubernetes/main/about/policies/target-merge.md
kristin-kronstain-brown: Good catch, added a paragraph: "A tie only happens when two policies share both the same specificity and the same field. You can attach multiple policies to the same target as long as each one sets a different field, or attaches at a different specificity level. For example, a `frontend` policy that sets `tls` at the Gateway level and another that sets `accessLog` using a listener `sectionName` don't tie, because they set different fields. A `frontend` policy that sets `tls` at the Gateway level and another that sets `tls` using `port` also don't tie, because `port` is more specific and wins for that field."

content/docs/kubernetes/main/about/policies/target-merge.md
kristin-kronstain-brown: I think it's correct. Claude checked it against `crates/agentgateway/src/store/policy.rs`, where Override locks the field once the less-specific policy sets it. But the name is genuinely confusing, so I added a note: "The less-specific policy is the one that overrides, not the one being overridden. A Gateway-level `traffic` policy with `inheritance: Override` locks its fields so that policies attached at more-specific points, such as routes or route rules, can't replace them."

content/docs/kubernetes/main/about/policies/target-merge.md
kristin-kronstain-brown: Fixed. Added a transition sentence ("For backend, precedence works the same way but with more levels...") so it's clear the example switches from frontend to backend instead of reading like an unexplained continuation.

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,36 @@ This field-level merge applies to all fields, including nested sub-fields. Each
...
+
+A tie only happens when two policies share both the same specificity and the same field. You can attach multiple policies to the same target as long as each one sets a different field, or attaches at a different specificity level. For example, a `frontend` policy that sets `tls` at the Gateway level and another that sets `accessLog` using a listener `sectionName` don't tie, because they set different fields. A `frontend` policy that sets `tls` at the Gateway level and another that sets `tls` using `port` also don't tie, because `port` is more specific and wins for that field.
+
+A tie is easiest to hit with `frontend`, because a `frontend` policy has at most three specificity levels: the Gateway, an optional `port`, and an optional listener `sectionName` for the fields where `sectionName` is allowed. Two policies that both target only the Gateway have no way to differentiate their specificity.
+
+To avoid a tie, set a given field or section in only one policy per target.
+
+## Merge strategy overrides {#strategy}
```
Nadine2016, line 83:
```suggestion
## Merge strategy overrides for traffic policies {#strategy}
```

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,36 @@ This field-level merge applies to all fields, including nested sub-fields. Each
...
+
+`AgentgatewayPolicy.spec.strategy.inheritance` changes how `traffic` policies merge. It's valid only on policies that set `traffic`. The Kubernetes API server rejects a policy that sets the `inheritance` block alongside the `frontend` or `backend` block. Frontend and backend policy merging always follows the `frontend` and `backend` precedence orders in the [Merge precedence](#merging-precedence) table and never uses `inheritance`.
+
```
Nadine2016, line 86: The following `inheritance` values are supported for `traffic` policies:

content/docs/kubernetes/main/about/policies/target-merge.md
```diff
@@ -47,12 +58,36 @@ This field-level merge applies to all fields, including nested sub-fields. Each
...
+| Value | Behavior |
+| -- | -- |
+| `Default` | The default value. Fields from more-specific attachment points, such as routes and route rules, can override fields from less-specific attachment points, such as gateways and listeners. Use this value to set a `traffic` default at the Gateway that specific routes can override. |
+| `Override` | Blocks `traffic` policies at more-specific attachment points from contributing to the effective policy. Use this value when a less-specific policy, such as one at the Gateway level, must stay authoritative for everything below it. |
+
+> [!NOTE]
```
Nadine2016, line 92: I would add this text to the override section as it only applies there

Nadine2016 (approved):

content/docs/kubernetes/main/about/policies/target-merge.md
kristin-kronstain-brown: Added

content/docs/kubernetes/main/about/policies/target-merge.md
kristin-kronstain-brown: Done

PR #884, opened 2026-08-11
docs: clarify virtual model failover requires health.eviction

guenhter:
Cross-priority failover only happens after targets are evicted, so document priority groups, health vs eviction, and update the examples accordingly.

-- 

AI did most of the work analysing the failover handling and wrote most of the text. I reviewed it myself.


content/docs/standalone/latest/llm/virtual-models.md
```diff
@@ -154,8 +156,38 @@ assert_models config-weighted.yaml '["gpt-4o-public","smart"]'
...
+Across priority groups, traffic moves to the next group only after every target in the current group is **evicted**. Lowering a health score alone is not enough to spill over to the next priority.
+
+#### Health vs eviction
+
+Configure health on the concrete `llm.models[]` entries that the virtual model targets (not on the virtual model itself).
+
+| Setting | What it does |
+| -- | -- |
+| No `health` policy | Unhealthy responses (by default, `5xx` or connection failures) still lower the endpoint health score used for within-group load balancing. Endpoints are never evicted, so traffic never fails over to the next priority. |
+| `health` without `eviction` | Same score-based weighting within a group. No eviction, so no cross-priority failover. |
```
artberger: Is this always the case? elsewhere we have these exceptions:
```suggestion
| `health` without `eviction` | Same score-based weighting within a group. No eviction, so no cross-priority failover. Targets are evicted only when agentgateway can derive an eviction duration from elsewhere: the `backoff` on a retry policy, or a `Retry-After` header on a 429 unhealthy response. |
```

content/docs/standalone/latest/llm/virtual-models.md
```diff
@@ -154,8 +156,38 @@ assert_models config-weighted.yaml '["gpt-4o-public","smart"]'
...
+| `health.eviction` | Removes an unhealthy endpoint from the active set for a backoff period. When every endpoint in a priority group is evicted, later requests use the next priority. |
+
+{{< callout type="warning" >}}
+`routing.failover` alone does **not** switch to a lower-priority target after errors. You must set `health.eviction` on the primary (and typically backup) concrete models. Without eviction, requests keep hitting the highest-priority group forever.
+{{< /callout >}}
```
artberger: (callouts are an older style we are moving away from)
```suggestion
> [!WARNING]
> Setting `routing.failover` alone does **not** switch to a lower-priority target after errors. You must set `health.eviction` on the primary (and typically backup) concrete models. Without eviction, requests keep hitting the highest-priority group forever.
```

content/docs/standalone/latest/llm/virtual-models.md
```diff
@@ -154,8 +156,38 @@ assert_models config-weighted.yaml '["gpt-4o-public","smart"]'
...
+
+Useful `health` fields:
+
+- `unhealthyExpression`: Optional CEL expression; `true` marks the response unhealthy. When unset, any `5xx`, non-zero gRPC status, or connection failure is unhealthy.
+- `eviction.duration`: Base time to keep an endpoint evicted. Defaults to `3s` when `eviction` is set but `duration` is omitted. Repeated evictions use multiplicative backoff.
```
artberger: We phrase this elsewhere as...
```suggestion
- `eviction.duration`: Base time to keep an endpoint evicted. When you omit `duration`, agentgateway uses the `Retry-After` value from a 429 response, then the `backoff` from a retry policy, and then a default of `3s`. Repeated evictions use multiplicative backoff, with no upper bound.
```

artberger (commented):
Thanks @guenhter , I had a couple phrasing questions.

content/docs/standalone/latest/llm/virtual-models.md
guenhter: You are completely wright. I double checked in the code and changed the description.

content/docs/standalone/latest/llm/virtual-models.md
guenhter: done

content/docs/standalone/latest/llm/virtual-models.md
guenhter: changed.

artberger (approved):
Thanks for revising!

PR #903, opened 2026-08-14
add dcr client reg to jwt guide

artberger:
Followup to #901
- updates the JWT guide to use DCR
- adds an authz section to the JWT guide


wkrause13:
This change makes the automated JWT test work, but it does not show the MCP UX I think we're trying to document.

The expected experience is:

1. A user adds the MCP server URL to a tool like Claude.
2. Claude registers itself automatically with Keycloak.
3. The user is redirected to Keycloak and signs in with their normal account.
4. Agentgateway checks the resulting token and grants access based on its audience and the user’s group membership.

The proposed change instead requires an administrator to create a registration token and register a special client first. It then sends the user’s username and password directly to Keycloak with `curl`. That is useful for non-interactive test automation, but it is not how Claude or another MCP client authenticates.

Requiring administrator-assisted client registration removes much of the value of DCR. Using the password grant also bypasses the browser-based login experience, including SSO, MFA, and normal Keycloak authentication policies.

If the password grant is required by CI, we should keep it in hidden test setup. The visible guide should demonstrate automatic client registration followed by the authorization-code flow with PKCE and a normal Keycloak login.

assets/agw-docs/pages/security/jwt-setup.md
```diff
@@ -48,14 +48,16 @@ Configure an {{< reuse "agw-docs/snippets/policy.md" >}} to validate JWTs using
...
+   | `audiences` | List of allowed audience values. The JWT's `aud` claim must contain at least one of these values. Omit the field to accept any audience.{{< version exclude-if="1.4.x,1.3.x,1.2.x,1.1.x,1.0.x" >}} An empty list also accepts any audience, and a non-empty list rejects a token that has no `aud` claim.{{< /version >}}<br><br>Example value: `["my-application"]` |
+   | `jwks.remote.jwksPath` | The path to the JWKS endpoint on the identity provider, relative to the backend root. This endpoint returns the public keys used to verify JWT signatures.<br><br>Example value: `/realms/master/protocol/openid-connect/certs` |
+   | `jwks.remote.cacheDuration` | How long to cache the JWKS keys locally. This reduces load on the identity provider and improves performance. Keys are automatically refreshed when the cache expires.<br><br>Example value: `5m` (5 minutes) |
```
Nadine2016:
```suggestion
   | `jwks.remote.cacheDuration` | How long to cache the JWKS keys locally. This setting reduces load on the identity provider and improves performance. Keys are automatically refreshed when the cache expires.<br><br>Example value: `5m` (5 minutes) |
```

assets/agw-docs/pages/security/jwt-setup.md
```diff
@@ -48,14 +48,16 @@ Configure an {{< reuse "agw-docs/snippets/policy.md" >}} to validate JWTs using
...
+   | `jwks.remote.backendRef` | Reference to the backend that hosts the identity provider. Agentgateway uses this to fetch the JWKS from the identity provider. For an in-cluster provider, reference a Kubernetes Service. For an external provider reached over TLS, reference an {{< reuse "/agw-docs/snippets/backend.md" >}} instead. See [External identity provider over TLS](#external-identity-provider-over-tls). <br><br>Example value: The Keycloak service |
```
Nadine2016:
```suggestion
   | `jwks.remote.backendRef` | Reference to the backend that hosts the identity provider. Agentgateway uses this value to fetch the JWKS keys from the identity provider. For an in-cluster provider, reference a Kubernetes Service. For an external provider that is reached over TLS, reference an {{< reuse "/agw-docs/snippets/backend.md" >}} instead. See [External identity provider over TLS](#external-identity-provider-over-tls). <br><br>Example value: The details of the Keycloak service |
```

Nadine2016: the example value looks weird

assets/agw-docs/pages/security/jwt-setup.md
```diff
@@ -92,21 +94,44 @@ Now that JWT authentication is configured, test the setup by obtaining a token f
    authentication failure: no bearer token found%  
    ```      
    
-2. Get an access token from Keycloak by using the password grant type.
+2. Register a client with Keycloak. The client uses dynamic client registration (DCR), so no administrator creates it, and Keycloak returns the client ID and secret that the next step uses.
+
+   > [!NOTE]
```
Nadine2016:
```suggestion
   > [!WARNING]
```

assets/agw-docs/pages/security/jwt-setup.md
```diff
@@ -141,6 +166,146 @@ Now that JWT authentication is configured, test the setup by obtaining a token f
    ```
   
 
+## Authorize requests by a JWT claim
+
+Authentication proves who sent the request. Authorization decides what that identity is allowed to do. After agentgateway validates the JWT, the claims are available to Common Expression Language (CEL) expressions through the `jwt` variable, so you can write access rules against them in the same {{< reuse "agw-docs/snippets/policy.md" >}}.
+
+This distinction matters when DCR is enabled. Registration lets a client obtain a token from the identity provider, but it does not grant that client access to your backends. Authorization is what decides which of those identities the gateway accepts.
```
Nadine2016:
```suggestion
This distinction matters when DCR is enabled. Registration lets a client obtain a token from the identity provider, but it does not grant that client access to your backends. Authorization is what decides what actions the client can perform.
```

assets/agw-docs/pages/security/jwt-setup.md
```diff
@@ -141,6 +166,146 @@ Now that JWT authentication is configured, test the setup by obtaining a token f
...
+
+The following example allows only the client that you registered, and denies every other identity.
```
Nadine2016: maybe move this sentence to step 1

Nadine2016 (approved):

assets/agw-docs/pages/security/jwt-setup.md
```diff
@@ -141,6 +166,146 @@ Now that JWT authentication is configured, test the setup by obtaining a token f
...
+
+1. Update the `jwt-auth-policy` to add an authorization rule.
```
artberger:
```suggestion
1. Update the `jwt-auth-policy` to add an authorization rule. The following example allows only the client that you registered, and denies every other identity.
```

PR #919, opened 2026-08-19
docs: document the admin config-resources API for managing API keys at runtime

erentorlak:
Documents the agentgateway admin API (`/api/config/resources/*`) for managing config resources such as API keys at runtime, without editing the config file. This endpoint was previously undocumented.

## What this adds

A new standalone "Operations" page describing the admin API:

- Resource kinds (dot-delimited, e.g. `llm.apiKey`)
- List / create (collection PUT, upsert) / update (single PUT, key preserved) / delete endpoints
- The `storage.mode: hybrid` requirement and DB persistence
- Security warning: the admin port (15000) has no authentication
- Propagation timing notes for create/delete
- Config-file vs admin-API keys comparison

Content stubs were added for both `latest` and `main`; the body lives in the shared asset `assets/agw-docs/pages/operations/config-resources-standalone.md`.

## Why it matters

The `llm.policies.apiKey.keys` config path only documents keys in the config file. With `config.storage.mode: hybrid`, keys are stored in the database and managed through this API (used by the Admin UI and by integrations that create/revoke keys at runtime). Without documentation, users either edit `config.yaml` (and miss hybrid mode) or reverse-engineer the endpoint.

## Notes / scope

- Docs-only change; no code, schema, or test changes.
- `test: skip` is used on the page stub, mirroring the existing `inspect-config` operations page, because the page documents a runtime API that requires a running hybrid-mode instance.

Related: https://github.com/agentgateway/agentgateway/issues/3085 (the "What is not stored" table should also note that API keys are DB-stored under hybrid mode).


artberger (approved):
Thanks for the contribution! fyi, I added some commits to align with style, some new topics, naming, and such.

PR #927, opened 2026-08-20
correct menu from grafana

linsun:
Can't find the Configurations, had to google it...

image


linsun:
Do i need to update main as well? not sure if it is auto synched?

artberger (approved):

PR #931, opened 2026-08-21
docs: invoice-grade attribution for Bedrock and Vertex (standalone plus Kubernetes)

Lyqed:
Added a new page under LLM → Cost controls in both the standalone and Kubernetes trees, linked from each section's index. It documents three things that shipped without docs: STS session tags and session name on assumeRole (agentgateway/agentgateway#2435, #2447, #2508), finalTransformations (agentgateway/agentgateway#2912) as the way to set Bedrock request metadata and Vertex billing labels, and the merge/replace postures discussed on agentgateway/agentgateway#2806. It also covers where each value lands (CloudTrail, invocation logs, Cost Explorer, GCP billing export) and the provider limits, to make it clear.

Examples were checked against main.


artberger (approved):
Thank you for the contribution! Just fyi, I will make a small followup PR to clean up some things related to style, formatting, etc.

Lyqed:
Thank you for organizing things up. 

per-request metadata attribution should probably move out out of cost-controls at some point.

PR #940, opened 2026-08-24
Update istio/agw ingress guide

jaellio:
- remove kgateway references
- match sytle of egress guide


artberger (approved):
Thank you so much!

PR #948, opened 2026-08-25
Model catalog and router additions

kristin-kronstain-brown:


content/docs/kubernetes/main/llm/models/about.md
```diff
@@ -92,6 +111,109 @@ spec:
         kind: HTTPRoute
 ```
 
+## Path-scoped models on an HTTPRoute
+
+A `Gateway` parent gives a listener one model router at the listener root. That is enough for a single set of models, but not when one listener must serve several independent sets. To create additional routers on the same listener, declare each one with an `HTTPRoute` and attach models to the route instead of the Gateway.
```
kristin-kronstain-brown:
```suggestion
A `Gateway` parent gives a listener one model router at the listener root. That setup is enough for a single set of models, but not when one listener must serve several independent sets. To create additional routers on the same listener, declare each one with an `HTTPRoute` and attach models to the route instead of the Gateway.
```

assets/agw-docs/snippets/model-catalog-tags.md
```diff
@@ -0,0 +1,53 @@
+A tag is a freeform string on a model entry that describes the model rather than prices it. Tags let one catalog carry model attributes next to cost data, so you can change how {{< reuse "agw-docs/snippets/agentgateway.md" >}} treats a model by editing the catalog instead of the gateway configuration.
```
Rachael-Graham: this actually may be perfectly correct grammar but it feels off for some reason??
```suggestion
A tag is a freeform string on a model entry that describes the model rather than pricing it. Tags let one catalog carry model attributes next to cost data, so you can change how {{< reuse "agw-docs/snippets/agentgateway.md" >}} treats a model by editing the catalog instead of the gateway configuration.
```

assets/agw-docs/snippets/model-catalog-tags.md
```diff
@@ -0,0 +1,53 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+### Chat format tags
+
+{{< reuse "agw-docs/snippets/agentgateway-capital.md" >}} interprets one family of tag values today: the request formats that a GitHub Copilot model accepts. Copilot exposes different endpoints for different models, so {{< reuse "agw-docs/snippets/agentgateway.md" >}} keeps a built-in list per model and converts the client's request into a format on that list. Tag a model to replace the built-in list, for example when GitHub changes which endpoints a model serves.
```
Rachael-Graham: > Tag a model to replace the built-in list, for example when GitHub changes which endpoints a model serves.

this reads funky to me but i dont have a good idea to replace/reword

assets/agw-docs/snippets/model-catalog-tags.md
```diff
@@ -0,0 +1,53 @@
...
+| `openai_responses` | OpenAI Responses |
+| `anthropic_messages` | Anthropic Messages |
+| `bedrock_converse` | Amazon Bedrock Converse |
+| `vertex_gemini` | Google Vertex AI Gemini |
+
+The tags apply as follows.
+
+- A model entry that carries at least one tag from this table replaces the built-in list for that model. Only the formats that you tag are accepted, so list every format that the model serves.
+- A model entry with no tags, or with only tags outside this table, keeps the built-in list.
+- {{< reuse "agw-docs/snippets/agentgateway-capital.md" >}} lowercases the requested model name before it looks up tags. Write model names in the catalog in lowercase, otherwise the lookup misses and the built-in list applies.
```
Rachael-Graham:
```suggestion
- {{< reuse "agw-docs/snippets/agentgateway-capital.md" >}} lowercases the requested model name before it looks up tags. Write model names in the catalog in lowercase; otherwise, the lookup misses, and the built-in list applies.
```

assets/agw-docs/snippets/model-catalog-tags.md
```diff
@@ -0,0 +1,53 @@
...
+- A client request in a format that the model does not accept fails with an unsupported conversion error that lists the accepted formats.
+
+These tags apply only to the `copilot` provider, which is available in standalone mode. Tag values outside this table are stored and merged, but {{< reuse "agw-docs/snippets/agentgateway.md" >}} does not act on them yet.
```
Rachael-Graham:
```suggestion
These tags apply only to the `copilot` provider, which is available in standalone mode. Tag values that are not listed in this table are stored and merged, but {{< reuse "agw-docs/snippets/agentgateway.md" >}} does not act on them yet.
```

content/docs/kubernetes/main/llm/models/about.md
```diff
@@ -92,6 +111,109 @@ spec:
...
+
+- `GET /tenant-a/v1/models` lists `gpt-5-mini` and no other models.
+- `POST /tenant-a/v1/chat/completions` with `"model": "gpt-5-mini"` serves the request.
+- Both paths require the basic authentication credentials that the policy sets.
+
+Agentgateway strips the `/tenant-a` prefix before it forwards the request, so the provider receives the standard `/v1/chat/completions` path.
+
+### Router scoping constraints
+
+- **A virtual model and the concrete models it selects must share one router.** A `weighted` or `conditional` virtual model resolves its targets by model name inside its own router's table. Attach the virtual model and its targets to the same parent, otherwise the request fails with `virtual_model_not_resolved`.
```
Rachael-Graham:
```suggestion
- **A virtual model and the concrete models it selects must share one router.** A `weighted` or `conditional` virtual model resolves its targets by model name inside its own router's table. Attach the virtual model and its targets to the same parent; otherwise, the request fails with `virtual_model_not_resolved`.
```

Rachael-Graham (approved):

assets/agw-docs/snippets/model-catalog-tags.md
kristin-kronstain-brown: Revised to:
> Tags override that list for a single model, which matters when GitHub changes which endpoints a model serves.

PR #960, opened 2026-08-26
OpenAI inline moderation for 1.5, both modes

kristin-kronstain-brown:
Adds provider-level inline moderation (openAI.moderation, upstream #2519) to the moderation guides in standalone and Kubernetes mode, and distinguishes it from the existing openAIModeration prompt guard, which has a near-identical name and a different mechanism.

Each page now opens with a comparison table, keeps the existing prompt guard content under a clearer heading, and adds an inline moderation section with requirements and limitations.

Verified against v1.5.0-beta.1 and live OpenAI:

- The standalone `llm:` config style cannot express this field, so the standalone example uses the `routes:` style.
- Casing is inverted between modes: `openAI`/`openai` for the provider key and lowercase/title case for the mode enum.
- `mode: block` does not stop flagged content. Reproduced with agentgateway removed from the path, on gpt-4o-mini, gpt-4o, gpt-4.1, and gpt-5, so both pages warn against relying on it and point at the prompt guard for a hard block.
- OpenAI requires `moderation.model`; a client value that omits it fails.
- `passthrough` and `detect` routes are not modified, so the parameter is not injected there.


content/docs/kubernetes/main/llm/guardrails/moderation.md
```diff
@@ -85,9 +102,148 @@ The OpenAI Moderation API detects potentially harmful content across categories
...
+
+2. Send a request to the LLM provider. 
+   {{< tabs >}}
+
+   {{% tab name="Cloud Provider LoadBalancer" %}}
+   ```sh
+   curl -i "$INGRESS_GW_ADDRESS/openai" \
+     -H "content-type: application/json" \
+     -d '{
+       "model": "gpt-4o-mini",
```
artberger: Not sure if we purposefully need to use 4o-mini here and throughout? We have a shortcode for the OpenAI model so that we can update it once for everywhere, but it's set to 3.5-turbo: `{{< reuse "agw-docs/snippets/openai-model." >}}`

content/docs/standalone/main/llm/prompt-guards/moderation.md
```diff
@@ -6,11 +6,28 @@ description: Use the OpenAI Moderation API as a prompt guard to screen LLM traff
 
 The OpenAI Moderation API detects potentially harmful content across categories including hate, harassment, self-harm, sexual content, and violence.
 
+## About the two moderation options
+
+Two separate features use the OpenAI Moderation API, and they have similar names. This page covers both. Read the following table to choose the one that fits your use case.
+
+| | Moderation prompt guard | Inline moderation |
```
artberger: The first column does not have a name, maybe something like Area?
```suggestion
| Area | Moderation prompt guard | Inline moderation |
```

content/docs/standalone/main/llm/prompt-guards/moderation.md
```diff
@@ -61,4 +78,126 @@ The OpenAI Moderation API detects potentially harmful content across categories
...
+
+* **The provider must be OpenAI.** The `moderation` field exists only on the `openAI` provider. Azure OpenAI is a separate provider and has no such field. To moderate traffic to any other provider, use the moderation prompt guard instead.
+* **The route type must be `completions` or `responses`.** Requests on a `passthrough` or `detect` route reach OpenAI unchanged, so the moderation parameter is not added.
+* **Clients keep their own `moderation` value when you omit the field.** If you do not configure `moderation`, a `moderation` value that a client sends passes through to OpenAI unchanged. OpenAI requires `moderation.model`, so a client value that omits it fails with `Missing required parameter: 'moderation.model'`. Configuring `moderation` on the backend avoids this, because agentgateway always sends a model.
+* **Moderation results reach the client only in OpenAI response formats.** A client that uses a different API format, such as the Anthropic Messages API, does not receive the moderation results.
+
+> [!NOTE]
+> The mode values are `block` and `score` in standalone mode. The Kubernetes mode documentation uses `Block` and `Score` in title case for the same fields.
```
artberger: Do we need this note since the doc sets are separate?

content/docs/kubernetes/main/llm/guardrails/moderation.md
```diff
@@ -85,9 +102,148 @@ The OpenAI Moderation API detects potentially harmful content across categories
...
+* **The route type must be `Completions` or `Responses`.** Requests on a `Passthrough` or `Detect` route reach OpenAI unchanged, so the moderation parameter is not added. The [OpenAI provider guide]({{< link-hextra path="/llm/providers/openai/" >}}) sets `"*": "Passthrough"` as a catch-all, so give the chat completions and responses paths an explicit route type, as shown in the preceding example.
+* **Clients keep their own `moderation` value when you omit the field.** If you do not configure `moderation`, a `moderation` value that a client sends passes through to OpenAI unchanged. OpenAI requires `moderation.model`, so a client value that omits it fails with `Missing required parameter: 'moderation.model'`. Configuring `moderation` on the backend avoids this, because agentgateway always sends a model.
+* **Moderation results reach the client only in OpenAI response formats.** A client that uses a different API format, such as the Anthropic Messages API, does not receive the moderation results.
+
+> [!NOTE]
+> The mode values are `Block` and `Score` in Kubernetes mode. The standalone mode documentation uses `block` and `score` in lowercase for the same fields.
```
artberger: Doe we need this note since the doc sets are separate?

content/docs/kubernetes/main/llm/guardrails/moderation.md
```diff
@@ -6,11 +6,28 @@ description: Detects potentially harmful content across categories including hat
 
 The OpenAI Moderation API detects potentially harmful content across categories including hate, harassment, self-harm, sexual content, and violence.
 
+## About the two moderation options
+
+Two separate features use the OpenAI Moderation API, and they have similar names. This page covers both. Read the following table to choose the one that fits your use case.
+
+| | Moderation prompt guard | Inline moderation |
```
artberger: First column is blank, maybe something like Area?
```suggestion
| Area | Moderation prompt guard | Inline moderation |
```

artberger (approved):

content/docs/kubernetes/main/llm/guardrails/moderation.md
kristin-kronstain-brown: Fixed

content/docs/standalone/main/llm/prompt-guards/moderation.md
kristin-kronstain-brown: Removed.

content/docs/kubernetes/main/llm/guardrails/moderation.md
kristin-kronstain-brown: Removed.

Issue #324, opened 2026-03-19
Agentgateway as a reverse proxy


artberger:
request from Lin during community meeting 3/19/2026

Document how we use agentgateway as a reverse proxy
- things like proxy API requests to a particular service inside the cluster
- for this path, set these headers, pass it to the service
- John has a pretty in-depth blog about using nginx for a local proxy over docker: https://blog.howardjohn.info/posts/agentgateway-at-home/
- This use case is common for demos

yashrajshuklaaa:
@artberger 
i'd like to work on this issue

PR #961, opened 2026-08-26
Standalone egress proxy guide for 1.5

kristin-kronstain-brown:
Documents agentgateway as an HTTP CONNECT forward proxy, covering tunnel mode (upstream #3014) and client authentication via Proxy-Authorization (upstream #3098) in one guide, since the two features interact.

Four patterns, each verified end to end against v1.5.0-beta.1:

- Allowlist destinations by SNI without decrypting traffic
- Require client authentication, with a hostname allowlist
- Terminate TLS for selected hostnames
- Inspect all HTTPS traffic with a dynamic CA

Scope notes:

- Standalone only. The controller derives tunnelProtocol from the Gateway listener protocol and can only return HBONE_GATEWAY or DIRECT, so there is no path to `connect` in Kubernetes mode, and neither tunnelProtocol, dynamicCa, nor BindMode appears in any CRD.
- Not covered by website issue #324, which is about a reverse proxy.

The finding that shaped the page: `tunnelProtocol: connect` handles the CONNECT request before route matching, so listener policies are not applied to it. Configuring basicAuth on a tunnel-mode bind silently authenticates nothing. Proxy authentication requires `frontendPolicies.connect.mode: route`, which also refuses a blocked destination with a 404 at the proxy instead of a deferred TLS error.


content/docs/standalone/main/configuration/egress-proxy.md
```diff
@@ -0,0 +1,400 @@
...
+
+A client selects the proxy through the standard environment variable that most tools read.
+
+```sh
+export HTTPS_PROXY=http://127.0.0.1:3000
+```
+
+The client then sends `CONNECT <host>:443` to agentgateway instead of connecting to the destination directly. What agentgateway does with that request depends on which of the two CONNECT handling modes you configure.
+
+| | Tunnel mode | Route mode |
```
artberger: First column is blank, maybe Area?
```suggestion
| Area | Tunnel mode | Route mode |
```

artberger (approved):

PR #967, opened 2026-08-27
Fix the doc-test linkVersion map so canonical version gates match

artberger:
_load_link_version_map() read params.sections.<section>.versions, a config shape docs-theme-extras removed in 0.2.2 in favour of one params.versions list whose entries carry a sections: tag. hugo.yaml moved with the theme, so params.sections.kubernetes has been an empty table and the map has been empty ever since.

An empty map means _evaluate_version_block only ever sees the linkVersion token, so include-if="1.5.x" and exclude-if="1.4.x" have been silent no-ops in every generated doc test. Two shared snippets are keyed on canonical versions and therefore rendered an empty version string:

  versions/n-patch.md              -> "" , so helm-version-flag.md emits
                                     "v" and helm fails with
                                     'improper constraint: "v"'
  snippets/install-agentgateway-binary.md
                                   -> VERSION="v", so the release URL 404s
                                     and curl writes "Not Found" to the
                                     binary path

That broke the install step of doc tests on PRs whose own content was fine.

Read params.versions instead, and warn on stderr when the map comes out empty rather than silently disabling every canonical-version gate.

Regenerating all 424 scripts changes 184 of them: every --version flag now resolves (v1.4.1 on latest, 0.0.0-latest-dev on main), and gated content is selected correctly for the first time -- the route-delegation ReferenceGrant step, which is excluded on 1.4.x, no longer leaks into the latest script, matching what Hugo renders on the page.

### DIFF
--- a/scripts/doc_test_extract.py
+++ b/scripts/doc_test_extract.py
@@ -3,6 +3,7 @@
 import argparse
 import json
 import re
+import sys
 import textwrap
 from dataclasses import dataclass, field
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


artberger:
closing as duplicate with some work that Kristin is doing

PR #1007, opened 2026-09-01
PullPolicy link fix

kristin-kronstain-brown:
Fixes a broken anchor in the generated API reference.

`crd-ref-docs` templates a `kubernetes.io` URL for every k8s type, so the `pullPolicy` field links to `#pullpolicy-v1-core`. That anchor doesn't exist — `PullPolicy` is a type alias with no section of its own — so the link drops readers at the top of a very long API page. Repointed to the concepts page, which is where the field is actually documented.

The rewrite is one `sed` added to the post-processing block that already runs in **Generate agentgateway API docs**, plus a guard that fails the run if the pattern stops matching, so a silently-stale regex can't quietly reintroduce the bad anchor.

## Why here and not elsewhere

- **Not in `knownTypes`.** `LinkForType` checks `IsKubeType` before `IsKnownType`, so any entry for a k8s type is silently ignored. An entry for `PullPolicy` used to live in the config and never took effect.
- **Not by editing `api-*.md`.** The reference docs regenerate nightly, so a hand-edit is reverted within a day.
- **Not a lychee exclusion.** The anchor is genuinely dead, not a checker artifact. Excluding it would hide a real defect rather than remove noise.

## Upstream

[elastic/crd-ref-docs#143](https://github.com/elastic/crd-ref-docs/pull/143) flips that ordering so `knownTypes` wins, which would make this workaround unnecessary. It's been open since March 2025 and needs a rebase. If it merges, we pick it up automatically (the workflow runs `@latest`) and can drop the `sed` in favor of a `knownTypes` entry.

### DIFF
--- a/.github/workflows/reference-docs.yaml
+++ b/.github/workflows/reference-docs.yaml
@@ -225,7 +225,19 @@ jobs:
           # underlying types (e.g. `map[string]CELExpression` -> #map[string]celexpression,
           # which never gets a heading), keeping the readable type text.
           sed -i.bak -E 's/_Underlying type:_ _\[(map\[[^]]*\][^]]*)\]\(#[^)]*\)_/_Underlying type:_ _\1_/g' "$TARGET_FILE"
+          # Repoint the PullPolicy link. crd-ref-docs templates a kubernetes.io URL for
+          # every k8s.io type, but PullPolicy is a type alias with no section of its own,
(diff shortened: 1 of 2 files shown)


kristin-kronstain-brown:
Moved into: https://github.com/agentgateway/website/pull/1004

PR #1039, opened 2026-09-15
Docs for per-key rate limits and certificate key and metrics

artberger:
Update main docs based off API changes in #1034:
- per-key, claim-based limits in the local rate limit and the virtual key guides
- certificate metrics
- read certificates from another key

Upstream PRs:
- https://github.com/agentgateway/agentgateway/pull/3351, https://github.com/agentgateway/agentgateway/issues/1909)
- https://github.com/agentgateway/agentgateway/pull/3419, https://github.com/agentgateway/agentgateway/issues/3418
- https://github.com/agentgateway/agentgateway/pull/3170, https://github.com/agentgateway/agentgateway/issues/3166


assets/agw-docs/pages/agentgateway/mcp/rate-limit.md
```diff
@@ -74,6 +74,7 @@ Review the following table for example use cases and configuration guidance.
 | Allow burst for session initialization | Add `burst` because each session needs several requests before the first tool call runs. |
 | Hard ceiling across all gateway traffic | {{< reuse "agw-docs/snippets/policy.md" >}} on `Gateway`, `local[].requests`. |
 | Per-tool rate limits (e.g. tighter for expensive tools) | Global rate limit + CEL descriptors extracting `body.method` and `body.params.name`. |
+{{% version exclude-if="1.0.x,1.1.x,1.2.x,1.3.x,1.4.x,1.5.x" %}}| Give each caller its own limit | Add `local[].key`, such as `jwt.sub`. See [Claim-level rate limits](#claim-level). |{{% /version %}}
```
kristin-kronstain-brown: **Blocking.** This is the markdown-leak failure in `Static + content checks`, and it breaks the table on every released version.

The gated row sits between two existing rows. On any version where the gate excludes it, the shortcode emits nothing and leaves a blank line, which terminates the table. The "Combine auth + rate limiting" row then renders as a literal line of pipes in a `<p>` below the table.

I built the branch locally and confirmed it. From `public/docs/kubernetes/latest/documentation/mcp/rate-limit/index.html`:

```html
</table> </div> <p>| Combine auth + rate limiting | Apply both <code>mcp.authentication</code> and <code>traffic.rateLimit</code> in the same AgentgatewayPolicy or use separate policies. |</p>
```

The same leak appears in `latest`, `2.2.x`, `1.4.x`, `1.3.x`, `1.2.x`, `1.1.x`, and `1.0.x` — seven pages, which is the leak count CI reports. Only `main` renders correctly.

The fix is to move the gated row to the end of the table, which is exactly what you did in `assets/agw-docs/pages/agentgateway/llm/rate-limit.md`, and that one renders clean in every version:

```suggestion
| Combine auth + rate limiting | Apply both `mcp.authentication` and `traffic.rateLimit` in the same {{< reuse "agw-docs/snippets/policy.md" >}} or use separate policies. |{{% version exclude-if="1.0.x,1.1.x,1.2.x,1.3.x,1.4.x,1.5.x" %}}
| Give each caller its own limit | Add `local[].key`, such as `jwt.sub`. See [Claim-level rate limits](#claim-level). |{{% /version %}}
```

That suggestion replaces this line and assumes you also delete the existing "Combine auth" line below it.

assets/agw-docs/pages/security/backendtls.md
```diff
@@ -527,6 +530,118 @@ YAMLTest -f - <<'EOF'
...
+
+   ```sh
+   kubectl get {{< reuse "agw-docs/snippets/policy.md" >}} nginx-backend-tls -n {{< reuse "agw-docs/snippets/namespace.md" >}} \
+     -o jsonpath='{.status.ancestors[0].conditions}' | jq .
+   ```
+
+   Example output:
+
+   ```
+   "message": "error extracting CA cert from Secret agentgateway-system/nginx-ca: missing key \"ca.crt\"",
```
kristin-kronstain-brown: **Blocking.** This example output names the wrong Secret and the wrong key, so it reads as though `key` were ignored.

The step above uses Secret `nginx-ca-custom-key` with `key: corporate-roots.pem`, but the message quotes `nginx-ca` and `ca.crt`. It looks copy-pasted from the previous section.

`cacert.Resolve` interpolates both values, so the real message names whichever key the reference actually asked for (`controller/pkg/agentgateway/cacert/ca.go` on upstream `main`):

```go
key := Key(ref)
...
caCRT, ok = secret.Data[key]
if !ok || len(caCRT) == 0 {
    return "", fmt.Errorf("error extracting CA cert from Secret %s: missing key %q", nn, key)
}
```

```suggestion
   "message": "error extracting CA cert from Secret agentgateway-system/nginx-ca-custom-key: missing key \"corporate-roots.pem\"",
```

assets/agw-docs/pages/security/rate-limit-claim-level.md
```diff
@@ -0,0 +1,166 @@
...
+    rateLimit:
+      local:
+      - requests: 2
+        unit: Minutes
+        key: jwt.sub
+```
+
+The following steps key the limit on a request header instead, so that you can see the behavior without an identity provider.
+
+1. Update the rate limit policy to key the limit on a user header.
```
kristin-kronstain-brown: **Should fix.** "Update the rate limit policy" applies a *new* policy. The `httpbin-rate-limit` policy from the previous section is never deleted, so the reader ends up with two AgentgatewayPolicies on the same HTTPRoute.

It does work — `store/binds.rs` merges route policy fields newest-first with last-writer-wins, so the later `httpbin-claim-level-rate-limit` overrides the earlier 3-per-second limit. But the reader is told they edited something they did not, and nothing on the page tells them why the old limit stopped applying.

Two options: reuse the name `httpbin-rate-limit` so the sentence is literally true, or keep the new name and say that the second policy overrides the first, with a pointer to the [Inheritance](#inheritance) section.

Worth noting that the doc test cannot catch this either. The `claim-level-rate-limit` path installs from scratch and never applies `httpbin-rate-limit`, so the test environment only ever has one policy, unlike a reader working the page in order.

assets/agw-docs/pages/security/rate-limit-claim-level.md
```diff
@@ -0,0 +1,166 @@
...
+* **How many buckets are kept**: Each rule keeps up to 65,536 buckets and drops the least recently used ones, which for that key is the same as never having been seen.
+* **Where buckets live**: Buckets are held in memory by a single proxy replica, so each replica enforces the limit separately. For a limit that is shared across replicas, use [global rate limiting](#global).
+* **Invalid expressions**: If the expression does not compile, the policy is accepted with the `PartiallyValid` reason, and the rest of the policy still applies. Check the policy status for the message `local rate limit key is not a valid CEL expression`.
+
+For the variables that you can read in a key, see [Variables and functions]({{< link-hextra path="/reference/cel/variables/" >}}).
+
+### Clean up the claim-level policy
```
kristin-kronstain-brown: **Should fix.** This cleanup heading lands in the middle of the host page, which already has a `## Cleanup` section at the end.

The rendered ToC of `rate-limit-http` shows both, adjacent to each other in reading order:

```
local        | Local rate limiting
claim-level  | Claim-level rate limits
clean-up-the-claim-level-policy | Clean up the claim-level policy
global       | Global rate limiting
conditional-execution | Conditional execution
cleanup      | Cleanup
```

Move the `kubectl delete` into the page's `## Cleanup` section instead, gated the same way the reuse is.

assets/agw-docs/standalone/virtual-keys.md
```diff
@@ -420,6 +420,96 @@ For more information about rate limiting configuration options, see [Rate limits
...
+   EOF
+   ```
+
+2. Restart agentgateway with the updated configuration.
+
+   ```sh
+   agentgateway -f config.yaml
+   ```
+
+3. Send requests with Alice's virtual key until the budget is exhausted.
```
kristin-kronstain-brown: **Should fix.** The command does not do what the step says, and the example output is wrong for the first request.

Step 4's configuration returns to the Step 1 form, which does not set `tokenize: true`, and step 2 just restarted the process, so the bucket is full at `maxTokens: 10`. The prompt count is not estimated, so the first request is admitted and returns a story. Only the second request sees the 260 response tokens charged and gets `rate limit exceeded`.

Step 2 splits this into two steps for exactly that reason. Either follow the same shape here, or wrap the `curl` in a loop so the step matches "until the budget is exhausted".

assets/agw-docs/standalone/virtual-keys.md
```diff
@@ -420,6 +420,96 @@ For more information about rate limiting configuration options, see [Rate limits
...
+       ]
+     }'
+   ```
+
+   Example output:
+   ```
+   rate limit exceeded
+   ```
+
+4. Send the same request with Bob's virtual key. Unlike in Step 2, the request succeeds, because Bob's key now has its own token bucket that Alice's requests do not draw from.
```
kristin-kronstain-brown: **Should fix.** "Unlike in Step 2" points at the wrong step. Bob never appears in Step 2, which only uses Alice's key. Bob is the key that gets blocked in Step 3, so that is the contrast a reader can actually make.

The section intro above has the same problem: "so that you can compare the results with Step 2".

This is also the only step in the section with no `Example output` block. A reader has no way to tell what success looks like here, and success is the entire point of the step.

assets/agw-docs/pages/security/rate-limit-claim-level.md
```diff
@@ -0,0 +1,166 @@
...
+     expect:
+       statusCode: 200
+   EOF
+   {{< /doc-test >}}
+
+Review the following behavior before you rely on a claim-level limit.
+
+* **Requests without a value**: Requests whose key is empty, or whose expression cannot be evaluated, such as a request with no `x-user` header in this example, all share one bucket. An empty key does not exempt a request from the limit. To apply a limit to only some requests, use [conditional policies]({{< link-hextra path="/documentation/about/policies/conditional-policies" >}}) instead.
+* **How many buckets are kept**: Each rule keeps up to 65,536 buckets and drops the least recently used ones, which for that key is the same as never having been seen.
+* **Where buckets live**: Buckets are held in memory by a single proxy replica, so each replica enforces the limit separately. For a limit that is shared across replicas, use [global rate limiting](#global).
```
kristin-kronstain-brown: **Should fix.** These three behavior bullets now appear in five places, and the wording has already drifted between them:

- here
- `assets/agw-docs/pages/agentgateway/llm/rate-limit.md:357-360`
- `assets/agw-docs/pages/agentgateway/mcp/rate-limit.md:237` (compressed into one paragraph)
- `content/docs/standalone/main/documentation/configuration/resiliency/rate-limits.md:447-448`
- `assets/agw-docs/standalone/virtual-keys.md:505-507`

Compare the bucket-bound sentence across three of them:

> Each rule keeps up to 65,536 buckets and drops the least recently used ones, which for that key is the same as never having been seen.

> Buckets are held in memory by a single proxy replica, and each rule keeps up to 65,536 of them, dropping the least recently used.

> Each rule keeps up to 65,536 buckets and drops the least recently used ones. For a virtual key that was dropped, the next request starts from a full bucket.

Same fact, three phrasings, and the number appears five times. When the bound changes upstream, four of these will be missed. This is a snippet — the k8s and standalone wording differ enough that it may need two, but not five.

assets/agw-docs/pages/security/rate-limit-claim-level.md
```diff
@@ -0,0 +1,166 @@
...
+       type: local
+     expect:
+       statusCode: 200
+   EOF
+   {{< /doc-test >}}
+
+Review the following behavior before you rely on a claim-level limit.
+
+* **Requests without a value**: Requests whose key is empty, or whose expression cannot be evaluated, such as a request with no `x-user` header in this example, all share one bucket. An empty key does not exempt a request from the limit. To apply a limit to only some requests, use [conditional policies]({{< link-hextra path="/documentation/about/policies/conditional-policies" >}}) instead.
+* **How many buckets are kept**: Each rule keeps up to 65,536 buckets and drops the least recently used ones, which for that key is the same as never having been seen.
```
kristin-kronstain-brown: **Nit.** "least recently used" is slightly stronger than the implementation supports. The keyed map is a `quick_cache::sync::Cache`, which is not a strict LRU.

Upstream says "least used" in both the code comment and the PR description:

> the cache holds 65536 buckets per rule and evicts the least used

Same wording appears in the other four copies of this bullet. Dropping "recently" costs nothing and stays true whichever eviction policy `quick_cache` uses.

content/docs/kubernetes/main/reference/cel/variables.md
```diff
@@ -29,6 +29,8 @@ Depending on the policy, different fields are accessible based on when in the re
 |------|-------------------|
 |Transformation| `source`, `request`, `jwt`, `mcp`, `extauthz`, `response`, `llm` |
 |Remote Rate Limit| `source`, `request`, `jwt` |
+|Local Rate Limit key (`requests`)| `source`, `request`, `jwt`, `apiKey` — the rule is checked before the LLM request is parsed, so a key cannot read `llm`. |
+|Local Rate Limit key (`tokens`)| `source`, `request`, `jwt`, `apiKey`, `llm` — the rule is charged after the LLM request is parsed, so a key can read fields such as `llm.requestModel`. |
```
kristin-kronstain-brown, line 33: **Nit.** Both new rows list `apiKey`, but the `Remote Rate Limit` row directly above does not — and `assets/agw-docs/snippets/rate-limit-token-budget-step.md` in this same PR uses `apiKey.user_id` in a remote rate limit descriptor.

So the table now reads as if a remote descriptor could not do what the budget guide tells the reader to do. The gap predates this PR, but the two new rows make it conspicuous. Worth adding `apiKey` to the Remote Rate Limit row while you are in here.

I confirmed the two new rows themselves are right: `api_key` auth applies at `proxy/httpproxy.rs:240` and the local rate limit at `:279`, same phase, so a `requests`-type key genuinely can read `apiKey`. The `llm` split matches `localratelimit.rs` and its `a_key_without_its_context_uses_the_shared_bucket` test.

assets/agw-docs/pages/security/backendtls.md
```diff
@@ -527,6 +530,118 @@ YAMLTest -f - <<'EOF'
...
+           key: corporate-roots.pem
+         sni: example.com
+         verifySubjectAltNames:
+         - example.com
+   EOF
+   ```
+
+3. Send another request to the NGINX server. You still get back a 200 HTTP response code, because the gateway now reads the CA certificate from the `corporate-roots.pem` key.
+
+   * **Cloud Provider LoadBalancer**
```
kristin-kronstain-brown, line 576: **Nit, and possibly deliberate.** This uses a bullet list where the earlier sections of the page use `{{< tabs >}}` (lines 268 and 372). It does match its immediate neighbour at line 471, so if the bullet form in the Secret sections is intentional, ignore this. Flagging only because the new `rate-limit-claim-level.md` in this same PR uses `{{< tabs >}}` for the identical LoadBalancer/port-forward pair.

assets/agw-docs/pages/agentgateway/llm/rate-limit.md
```diff
@@ -308,6 +310,56 @@ EOF
...
+{{< version exclude-if="1.0.x,1.1.x,1.2.x,1.3.x,1.4.x,1.5.x" >}}
+## Claim-level budget limits {#claim-level}
+
+Create claim-level budget limits with JWT authentication and CEL expressions.
+
+A token budget without a `key` field is shared by every request on the target, so a single caller can spend the budget for everyone. Set `key` to a CEL expression to give each distinct value its own token bucket with the limits of that rule. The CEL expression checks against claims in a JWT. This way, per-user, per-team, and per-model budgets do not require an external rate limit service.
+
+Before you begin, create a [JWT authentication policy]({{< link-hextra path="/documentation/security/jwt/" >}}) that targets the same backend that you want to rate limit, such as an OpenAI HTTPRoute in the following example. To key on a virtual key instead, use `apiKey.user_id`, as described in the [virtual keys guide]({{< link-hextra path="/documentation/llm/cost-controls/virtual-keys/" >}}).
+
+The following policy sets up the following budgets:
```
kristin-kronstain-brown: **Nit.** Trailing whitespace, and "The following policy sets up the following budgets".

```suggestion
The following policy sets up these budgets:
```

kristin-kronstain-brown (approved):

PR #1047, opened 2026-09-17
Document the 32 MiB LLM body buffer default

kristin-kronstain-brown:
## What changed

Three pages described the gateway body buffer as defaulting to a flat 2 MiB. As of agentgateway/agentgateway#3408, requests that are routed to an LLM backend get a larger default of 32 MiB, so a single large-context prompt is not rejected by a limit tuned for general traffic.

| Page | Change |
| ---- | ------ |
| `assets/agw-docs/pages/traffic-management/buffering.md` | Two things. The "About buffer limits" paragraph now states both defaults, where the larger one starts applying, and that setting `maxBufferSize` yourself replaces both. A new "Choose a buffer limit" section then gives the guidance the page never had, and absorbs the edge-proxy sentence that used to sit at the end of the intro paragraph. |
| `content/docs/kubernetes/main/documentation/traffic-management/extproc.md` | The ExtProc buffering note now carries the LLM caveat. |
| `content/docs/standalone/main/documentation/configuration/traffic-management/extproc.md` | The same, with the standalone field name. |

The `2mb` in the intro paragraph is also corrected to `2 MiB`, to match the unit the rest of the page and both `extproc.md` notes use.

## The 32 MiB default has a boundary, and it is the easy thing to get wrong

The larger default is not "LLM requests get 32 MiB". It is applied when the request is routed to an LLM backend, so anything buffering earlier in the request path still uses 2 MiB. A policy running in the `PreRouting` phase is the case to watch, because it runs before route selection and never reaches the larger limit.

This is stated in #3408 itself ("a pre-route policy buffering would still be 2mb") and confirmed in the diff: the limit is inserted inside `make_backend_call` (`crates/agentgateway/src/proxy/httpproxy.rs`), on both the `Backend::LLMRouter` branch and the `llm_provider` branch. Both run after routing, so nothing earlier reaches them.

## Why this class of change is easy to miss

#3408 did not move a default from one number to another. It removed the schema default outright: `max_buffer_size` went from `usize` with `#[serde(default)]` to `Option<usize>`, and `"default": 2097152` was deleted from the schema. The default now lives in code and depends on where in the request path you are.

That leaves a page reading as correct while being wrong, because no field is added or renamed to signal it.

The other claims are taken from the source rather than the title:

- `llm::DEFAULT_BUFFER_LIMIT` is `32 * 1024 * 1024` (`crates/agentgateway/src/llm/mod.rs`), against `http::DEFAULT_BUFFER_LIMIT` of `2_097_152` (`crates/http/src/lib.rs`).
- It is applied through `get_or_insert_with`, so it fires only when the field is unset. That is what supports the sentence about an explicit value replacing both defaults.

## The version gating is the part worth reviewing

`buffering.md` is shared prose reused by **seven** Kubernetes version trees (1.0.x through 1.4.x, latest, and main). Neither `2.2.x` nor either standalone tree reads it, so the kubernetes-mode link in the new section is safe. #3408 is not released yet, and `hugo.yaml` maps `main` to 1.6.x and `latest` to 1.5.x, so stating the new default unconditionally would have published it into six trees where it is false.

There are **two** gates, not one, and they exclude different sets:

| Gated content | `exclude-if` | Intent |
| ------------- | ------------ | ------ |
| The 32 MiB paragraph, the defaults table, the `PreRouting` example, and the "replaces both defaults" sentence | `1.0.x,1.1.x,1.2.x,1.3.x,1.4.x,1.5.x,2.2.x` | main only, because #3408 is unreleased |
| The "Keep the Gateway limit conservative" bullet, which links to the route-level buffer policy | `1.0.x,1.1.x,1.2.x,2.2.x` | The trees where `buffer.md` exists, so the link is not written into a tree that has no such page |

Everything else in the new "Choose a buffer limit" section is deliberately ungated. That guidance is true of every version, and the relocated edge-proxy sentence is pre-existing prose, which should not pick up a gate it never had.

The two `extproc.md` files need no gate, because they are per-version pages and only the `main` copies are edited here. The `latest` and 1.4.x copies are deliberately untouched.

### What the build actually renders

Verified by grepping the rendered HTML per tree, not by reading the shortcode. Each marker is body text rather than a heading, because a heading also appears in the table of contents and in the copy-as-markdown script block.

| Tree | 32 MiB block | "Choose a buffer limit" | Route-level bullet |
| ---- | ------------ | ----------------------- | ------------------ |
| `kubernetes/main` | present | present | present |
| `kubernetes/latest` | absent | present | present |
| `kubernetes/1.4.x` | absent | present | **absent** |
| `kubernetes/1.3.x` | absent | present | **absent** |
| `kubernetes/1.2.x` and older | absent | present | absent |
| `standalone/main` | present (from `extproc.md`) | n/a | n/a |
| `standalone/latest` | absent | n/a | n/a |

The gated-out versions read cleanly, with no double space where the sentences were removed.

> [!IMPORTANT]
> The two bolded rows are not what the second gate asks for, and this is worth knowing before anyone writes another gate on this file. `layouts/_shortcodes/version.html` resolves the page's version by matching a URL segment against an entry in `params.versions`. That list holds only 1.6.x/main and 1.5.x/latest, so on 1.0.x through 1.4.x nothing resolves, no entry is reached, and the shortcode emits nothing at all. Every `version` block on an unregistered tree renders empty, whether the condition is `include-if` or `exclude-if` and whatever versions it names.
>
> The outcome here is safe, because the bullet only fails to appear on two trees where it would have been correct rather than appearing on trees where it would be wrong. It is a silent gap rather than a silent error. But the second gate's `exclude-if` list is doing nothing: the same rendering would follow from any list, and 1.3.x and 1.4.x lose a true sentence for a reason that has nothing to do with them.
>
> Two ways forward, neither of which belongs in this pull request. Register the frozen trees in `params.versions` so gates evaluate on them, or make the shortcode fall back to the directory name when no entry matches. Both are `docs-theme-extras` and `hugo.yaml` changes with a much wider blast radius than one page.

## Deliberately not changed

`buffer.md` in both modes mentions `maxBufferSize` only as a cross-reference and states no default, so it needs no edit. The 1.4.x release note says 2 MB, which is correct as a record of that release.

The generated configuration schema page still renders `Default 2097152` for `frontendPolicies.http.maxBufferSize`, a default the schema no longer carries. That file is generator-owned, so it is out of scope here and is worth checking after the next reference regeneration.


danehans:
https://github.com/agentgateway/agentgateway/pull/3408 states:

> Note: this applies once we hit the LLM traffic path; a pre-route policy
buffering would still be 2mb

I reviewed the rendered https://f9694ea6.agentproxy.pages.dev/docs/kubernetes/main/documentation/traffic-management/buffering/ and feel that this note is not being surfaced. Should an example be added?

kristin-kronstain-brown:
@danehans Thanks for looking at this so quickly! Would you review another draft? I addressed the missing note and created a section for Choosing a buffer limit.
https://kkb-buffering.agentproxy.pages.dev/docs/kubernetes/main/documentation/traffic-management/buffering/

danehans (approved):
The updated doc looks good.

danehans:
The Links CI check needs to be resolved.

kristin-kronstain-brown:
Fixed links here: https://github.com/agentgateway/website/pull/1051/files

PR #1048, opened 2026-09-17
Document responseIdleTimeout and timeouts on the simplified llm section

kristin-kronstain-brown:
Covers two upstream changes that land on the same table.

- agentgateway/agentgateway#3310 — adds `responseIdleTimeout`
- agentgateway/agentgateway#3366 — makes the timeout policy available on the simplified `llm:` section

## What changed

"Route Timeouts" listed two timeouts and now lists three, adding `responseIdleTimeout`. The two existing rows each gained a clause: both `requestTimeout` and `backendRequestTimeout` stop applying once the response headers arrive, which the page did not say and which is the whole reason the new field exists.

A **Simplified (LLM)** tab joins the existing Simplified (MCP) and Routing-based tabs, because #3366 makes `llm.policies.timeout` valid.

## Why the third timeout is needed

Because both existing timeouts stop at the response headers, neither places any bound on how long a response *body* may take, and neither can tell a stalled stream from a slow one. `responseIdleTimeout` restarts its window on every body frame, so it bounds the gap between frames rather than the total duration. That terminates a backend that stops producing data mid-stream without capping a legitimately long response.

This matters most for streaming responses, which are expected to run for a long time and were previously unbounded in practice.

## Verified against a build from main

Every field name here was checked by running `--validate-only` against `cr.agentgateway.dev/agentgateway:latest-dev`, which is the same nightly build the doc tests on the `main` tree install. All four configs on the page validate, including the two new fields.

The check worth recording is the negative control. Feeding a bogus field under `llm.policies.timeout` returns:

```
unknown field `notARealTimeout`, expected one of
`requestTimeout`, `backendRequestTimeout`, `responseIdleTimeout`
```

That names the accepted set outright, which settles two things: `llm.policies.timeout` takes the full timeout policy rather than a subset, and `responseIdleTimeout` applies there too. #3366's own schema shows only the first two fields, because it merged about two hours before #3310, so the combined surface is only visible on a build that carries both.

The page's doc test now covers all three forms instead of two, and asserts `responseIdleTimeout` on both the routing-based and simplified LLM configs. It is the newest of the three timeouts, so a rename upstream would otherwise reach the page as prose nobody can run.

Not covered, and called out in the test comments: that requests actually time out at runtime, and that the idle window genuinely restarts per body frame. Both need a backend that stalls mid-response, which no fixture here provides.

## The field name is different in Kubernetes

For anyone extending this to the Kubernetes tree: the field is `responseIdleTimeout` in standalone and `responseIdle` in the Kubernetes API. These rows must not be copied across unchanged.

## Deliberately not changed

`content/docs/kubernetes/main/documentation/resiliency/timeouts/idle.md` looks like the matching page and is not. It is titled "Idle timeouts" and covers terminating inactive **HTTP/1 connections**, a different subject from gaps between response body frames.

The Kubernetes tree has no page covering `responseIdle` today. Adding one is a separate change, and needs the Kubernetes field name rather than this one.


content/docs/standalone/main/documentation/configuration/resiliency/timeouts.md
```diff
@@ -20,12 +20,15 @@ Request {{< gloss "Timeout" >}}timeouts{{< /gloss >}} allow returning an error f
 
 ## Route Timeouts
 
-You can configure two types of timeouts on a route.
+You can configure three types of timeouts on a route.
```
kristin-kronstain-brown:
```suggestion
You can configure these types of timeouts on a route.
```

content/docs/standalone/main/documentation/configuration/resiliency/timeouts.md
```diff
@@ -20,12 +20,15 @@ Request {{< gloss "Timeout" >}}timeouts{{< /gloss >}} allow returning an error f
...
 
 |Timeout|Description|
 |-|-|
-|`requestTimeout`|The time from the start of an incoming request, until the end of the response headers is received. Note if there are retries, this includes the total time across retries.|
-|`backendRequestTimeout`|The time from the start of a request to a backend, until the end of the response headers are completed. Note this is per-request, so with retries this is a per-retry timeout.|
+|`requestTimeout`|The time from the start of an incoming request, until the end of the response headers is received. Note if there are retries, this includes the total time across retries. The response body is not included, so use `responseIdleTimeout` to bound gaps between body frames.|
```
kristin-kronstain-brown:
```suggestion
|`requestTimeout`|The time from the start of an incoming request, until the end of the response headers is received. Note if there are retries, this time includes the total time across retries. The response body is not included, so use `responseIdleTimeout` to bound gaps between body frames.|
```

content/docs/standalone/main/documentation/configuration/resiliency/timeouts.md
```diff
@@ -20,12 +20,15 @@ Request {{< gloss "Timeout" >}}timeouts{{< /gloss >}} allow returning an error f
...
+|`backendRequestTimeout`|The time from the start of a request to a backend, until the end of the response headers are completed. Note this is per-request, so with retries this is a per-retry timeout. Like `requestTimeout`, this stops applying once the response headers arrive.|
```
kristin-kronstain-brown:
```suggestion
|`backendRequestTimeout`|The time from the start of a request to a backend, until the end of the response headers are completed. Note this time is per-request, so with retries this time is a per-retry timeout. Like `requestTimeout`, this retry process stops applying once the response headers arrive.|
```

content/docs/standalone/main/documentation/configuration/resiliency/timeouts.md
```diff
@@ -20,12 +20,15 @@ Request {{< gloss "Timeout" >}}timeouts{{< /gloss >}} allow returning an error f
...
+|`responseIdleTimeout`|The maximum time the response body may go without producing data. The window restarts on every body frame, so this bounds the gap between frames rather than the total time a response may take. Use it to terminate a backend that stalls mid-stream, without capping how long a legitimately long response may run. The timeout is disabled when the field is unset or set to zero, and it never applies to responses that switch protocols, so upgraded WebSocket and CONNECT tunnels are not terminated by it.|
```
kristin-kronstain-brown:
```suggestion
|`responseIdleTimeout`|The maximum time the response body might go without producing data. The window restarts on every body frame, so this range bounds the gap between frames rather than the total time a response might take. Use it to terminate a backend that stalls mid-stream, without capping how long a legitimately long response may run. The timeout is disabled when the field is unset or set to zero, and it never applies to responses that switch protocols, so upgraded WebSocket and CONNECT tunnels are not terminated by it.|
```

content/docs/standalone/main/documentation/configuration/resiliency/timeouts.md
```diff
@@ -20,12 +20,15 @@ Request {{< gloss "Timeout" >}}timeouts{{< /gloss >}} allow returning an error f
...
+
+Because `requestTimeout` and `backendRequestTimeout` both stop at the response headers, neither one places any bound on how long a response body may take, and neither can tell a stalled stream from a slow one. That gap is what `responseIdleTimeout` covers, which matters most for streaming responses that are expected to run for a long time.
```
kristin-kronstain-brown:
```suggestion
Because `requestTimeout` and `backendRequestTimeout` both stop at the response headers, neither one places any bound on how long a response body might take, and neither can tell a stalled stream from a slow one. That gap is what `responseIdleTimeout` covers, which matters most for streaming responses that are expected to run for a long time.
```

content/docs/standalone/main/documentation/configuration/resiliency/timeouts.md
```diff
@@ -20,12 +20,15 @@ Request {{< gloss "Timeout" >}}timeouts{{< /gloss >}} allow returning an error f
 
 ## Route Timeouts
 
-You can configure two types of timeouts on a route.
+You can configure these types of timeouts on a route.
```
Nadine2016, line 23: does this require a release note?

content/docs/standalone/main/documentation/configuration/resiliency/timeouts.md
```diff
@@ -20,12 +20,15 @@ Request {{< gloss "Timeout" >}}timeouts{{< /gloss >}} allow returning an error f
...
+|`requestTimeout`|The time from the start of an incoming request, until the end of the response headers is received. Note if there are retries, this time includes the total time across retries. The response body is not included, so use `responseIdleTimeout` to bound gaps between body frames.|
+|`backendRequestTimeout`|The time from the start of a request to a backend, until the end of the response headers are completed. Note this time is per-request, so with retries this time is a per-retry timeout. Like `requestTimeout`, this retry process stops applying once the response headers arrive.|
+|`responseIdleTimeout`|The maximum time the response body might go without producing data. The window restarts on every body frame, so this range bounds the gap between frames rather than the total time a response might take. Use it to terminate a backend that stalls mid-stream, without capping how long a legitimately long response may run. The timeout is disabled when the field is unset or set to zero, and it never applies to responses that switch protocols, so upgraded WebSocket and CONNECT tunnels are not terminated by it.|
```
Nadine2016, line 29:
```suggestion
|`responseIdleTimeout`|The maximum time the response body can go without producing data. The window restarts on every body frame, so this range bounds the gap between frames rather than the total time a response might take. Use this setting to terminate a backend that stalls mid-stream, without capping how long a legitimately long response might run. The timeout is disabled when the field is unset or set to zero, and it never applies to responses that switch protocols, so upgraded WebSocket and CONNECT tunnels are not terminated by it.|
```

content/docs/standalone/main/documentation/configuration/resiliency/timeouts.md
```diff
@@ -20,12 +20,15 @@ Request {{< gloss "Timeout" >}}timeouts{{< /gloss >}} allow returning an error f
...
+
+Because `requestTimeout` and `backendRequestTimeout` both stop at the response headers, neither one places any bound on how long a response body might take, and neither can tell a stalled stream from a slow one. That gap is what `responseIdleTimeout` covers, which matters most for streaming responses that are expected to run for a long time.
```
Nadine2016, line 31:
```suggestion
Because requestTimeout and backendRequestTimeout both stop measuring elapsed time once the response headers arrive, neither one places any bound on how long a response body might take, and neither can differentiate a stalled stream from a slow one. The responseIdleTimeout covers this gap by limiting the time that can pass between response body chunks, which matters most for streaming responses that are expected to run for a long time.
```

Nadine2016 (approved):

content/docs/standalone/main/release-notes/release-notes.md
```diff
@@ -41,6 +42,17 @@ The two sources also name some providers differently, and they disagree about wh
...
 ## 🌟 New features {#v16-new-features}
 
+### Resiliency {#v16-features-resiliency}
+
+#### `responseIdleTimeout` and simplified LLM timeouts {#v16-response-idle-timeout}
+
+<!-- ref: https://github.com/agentgateway/agentgateway/pull/3310 -->
+<!-- ref: https://github.com/agentgateway/agentgateway/pull/3366 -->
+
+A new `responseIdleTimeout` field bounds the gap between response body frames, rather than the total response duration. `requestTimeout` and `backendRequestTimeout` both stop measuring elapsed time once the response headers arrive, so neither one can terminate a backend that stalls mid-stream without also capping how long a legitimately long response is allowed to run. `responseIdleTimeout` restarts its window on every body frame, is disabled when the field is unset or set to zero, and never applies to responses that switch protocols, so upgraded WebSocket and CONNECT tunnels are unaffected. The timeout policy is also now available on the simplified `llm:` section, alongside the existing routing-based and simplified MCP forms.
```
kristin-kronstain-brown:
```suggestion
A new `responseIdleTimeout` field bounds the gap between response body frames, rather than the total response duration. `requestTimeout` and `backendRequestTimeout` both stop measuring elapsed time after the response headers arrive, so neither one can terminate a backend that stalls mid-stream without also capping how long a legitimately long response is allowed to run. `responseIdleTimeout` restarts its window on every body frame, is disabled when the field is unset or set to zero, and never applies to responses that switch protocols, so upgraded WebSocket and CONNECT tunnels are unaffected. The timeout policy is also now available on the simplified `llm:` section, alongside the existing routing-based and simplified MCP forms.
```

content/docs/standalone/main/documentation/configuration/resiliency/timeouts.md
kristin-kronstain-brown: Added.

PR #1050, opened 2026-09-17
docs: add inference benchmarks for Kubernetes and standalone

danehans:
## Company

Not applicable — documentation-only change.

## Contribution summary

- Add inference benchmark pages for Kubernetes and standalone modes in `latest` and `main`.
- Publish the Qwen3-32B/H100 campaign from agentgateway/benchmarks#3.

## Checklist

Logo file and listing requirements: not applicable.

- [x] Built the staged changes in an isolated directory with Hugo.
- [x] Verified summary and per-rate measurements against the published reports.
- [x] Checked benchmark pages, overview links, chart assets, and section ordering in both modes and versions.

## Preview

Reviewed the pages locally.


danehans:
@artberger PTAL

danehans:
@kristin-kronstain-brown PTAL

kristin-kronstain-brown:
Thank you for this, @danehans. The numbers check out: I compared every table value against `metrics.csv` at `522fc04` and they all match, and the reuse structure across both modes and versions is clean.

> **Blocking**
> 
> `environment.md` line 3: "It uses no EPP or scoring" reads as if the campaign itself uses no EPP, which contradicts the next paragraph. Suggested:
> 
> ```suggestion
> This {{< gloss "Benchmark campaign" >}}campaign{{< /gloss >}} compares EPP-based agentgateway routing with a plain Kubernetes Service across eight vLLM model servers. The Kubernetes Service baseline uses no [EPP](https://llm-d.ai/docs/architecture/core/router#llm-d-endpoint-picker-epp) or [scoring](https://llm-d.ai/docs/architecture/core/router/epp/configuration#plugins) to route inference requests to a model server endpoint.
> ```
> 
> **Questions only you can answer**
> 
> 1. The summary compares at 60 requests/s, where the Service baseline has been saturated at about 6.7 requests/s since 20 QPS. At 10 QPS, the Service has slightly higher output throughput (5,113 vs 4,739). Can you add one sentence that says where the results diverge, so that the TTFT reduction reads as the effect of saturation?
> 2. `failure_rate` is 0.0 for every stage in both CSVs. Can the latency caveat state that no requests failed, instead of only asking readers to check?
> 3. The chart legends say "k8s service (RR)". How does the baseline Service balance load? The prose should match the legend.
> 
> **Smaller items** 
> 
> - Pin the benchmark links to `522fc04` instead of `main`, as the README describes.
> - Move `benchmarking.md` into `inference/`, so that the generated cards list it. You can then remove `disableCards`, the manual card grid, and the `## Benchmarking` section from the four `_index.md` files. Weight 47 also ties with `realtime.md` and `virtual-models.md`.
> - Replace `{{< callout >}}` with GitHub alerts (`> [!NOTE]`, `> [!WARNING]`), which this repo uses elsewhere.
> - Nit: remove "intelligently" in `inference.md`, and the `n/a` sentence, because no `n/a` values appear in the tables.

PR #1055, opened 2026-09-18
Updates from API changes

kristin-kronstain-brown:
Doc updates from the API changes updated in the reference: https://github.com/agentgateway/website/pull/1052


assets/agw-docs/snippets/model-catalog-tags.md
```diff
@@ -35,7 +35,43 @@ The tags apply as follows.
...
 - A client request in a format that the model does not accept fails with an unsupported conversion error that lists the accepted formats.
 
-These tags apply only to the `copilot` provider, which is available in standalone mode. Tag values that are not listed in this table are stored and merged, but {{< reuse "agw-docs/snippets/agentgateway.md" >}} does not act on them yet.
+<!-- Bedrock tag support is new in 1.6. The two branches below list every
+     version 1.5 and older, so they stay a complete pair: one branch matches
+     any release in the list, the other matches everything else. Do NOT
+     shorten either list to just "1.5.x" — that reads as "every version except
+     1.5", which shows the 1.6 text on 1.4.x and older. Adding "main" to the
+     exclude side would be wrong for the opposite reason: it would need editing
+     at every release. As written, a new release needs no edit here. -->
```
artberger: Are these comments needed?

assets/agw-docs/snippets/model-catalog-tags.md
```diff
@@ -35,7 +35,43 @@ The tags apply as follows.
...
+{{< version include-if="1.0.x,1.1.x,1.2.x,1.3.x,1.4.x,1.5.x,2.2.x" >}}These tags apply only to the `copilot` provider, which is available in standalone mode. Tag values that are not listed in this table are stored and merged, but {{< reuse "agw-docs/snippets/agentgateway.md" >}} does not act on them yet.{{< /version >}}{{< version exclude-if="1.0.x,1.1.x,1.2.x,1.3.x,1.4.x,1.5.x,2.2.x" >}}These tags apply to the `copilot` provider, which is available in standalone mode, and to Amazon Bedrock. The `aws-bedrock-mantle` import source sets them on Bedrock models for you.{{< /version >}}
+
+{{% version exclude-if="1.0.x,1.1.x,1.2.x,1.3.x,1.4.x,1.5.x,2.2.x" %}}
+On Bedrock, the tags are narrower than the rules above, because they are read only after the request picks an endpoint.
```
artberger, line 46: Do the rules above not apply to Bedrock then? and should this be under the next H3 that is `Bedrock tags`?

assets/agw-docs/standalone/cost-catalog.md
```diff
@@ -45,27 +45,43 @@ For general LLM telemetry setup, see [Observe traffic]({{< link-hextra path="/do
...
+
+{{% version exclude-if="1.0.x,1.1.x,1.2.x,1.3.x,1.4.x,1.5.x,2.2.x" %}}
+| Source | What it contributes |
+|--------|---------------------|
+| `models.dev` | Rates for every provider that the proxy supports, from [models.dev](https://models.dev). |
+| `aws-bedrock-mantle` | Tags for Amazon Bedrock models only, read from the AWS model cards. This source contributes no rates. The tags record which endpoint serves a model, `runtime` or `mantle`, and which request formats the Mantle endpoint accepts. |
+| `github` | The curated catalog that the agentgateway project publishes at [agentgateway.dev/model-catalog](https://agentgateway.dev/model-catalog), which covers the models that the agentgateway project tracks rather than everything that models.dev lists. Not imported by default. |
```
artberger, line 61: Are there other sources? For example, later we have `anthropic,mistral,openai` So I am not sure what this table is of, maybe only of the custom providers that agentgateway provides, not counting the llm providers like anthropic etc?

content/docs/kubernetes/main/documentation/llm/cost-controls/costs.md
```diff
@@ -32,14 +32,20 @@ Prepare a catalog by creating your own JSON file or using the `agctl catalog imp
 
 Use `agctl catalog import` to generate a catalog JSON file, then load it into a ConfigMap.
 
-1. Generate a catalog from a supported source. By default, `agctl catalog import` imports every provider in the `github` source, which is the curated catalog that the agentgateway project publishes at [agentgateway.dev/model-catalog](https://agentgateway.dev/model-catalog). To import from [models.dev](https://models.dev) instead, pass `--source models.dev`, which imports every provider that the proxy supports. To import only a subset of providers, pass a comma-separated list to `--providers`.
+1. Generate a catalog from one or more supported sources. The `--source` flag takes a comma-separated list, and the sources merge in the order that you list them, so a later source overlays an earlier one. By default, `agctl catalog import` imports `models.dev,aws-bedrock-mantle`, which prices every provider that the proxy supports and then tags the Amazon Bedrock models. To import only a subset of providers, pass a comma-separated list to `--providers`.
+
+   | Source | What it contributes |
```
artberger, line 37: Similar question about this table

assets/agw-docs/standalone/cost-catalog.md
artberger: Reading a little later, looks like the anthropic etc providers are actually provider IDs of `models.dev` (and the other "providers" here might have their own provider IDs that could also be passed in)

content/docs/standalone/main/integrations/llm/providers/bedrock.md
```diff
@@ -195,8 +203,54 @@ See [here](../anthropic/#use-claude-platform-on-aws) for connect to [Claude Plat
...
+| Value | Endpoint selection |
+|-------|--------------------|
+| `runtimePreferred` | Use Runtime, except for a model tagged `mantle` but not `runtime`. This value is the default. |
+| `mantlePreferred` | Use Mantle, except for a model tagged `runtime` but not `mantle`. |
+| `runtimeOnly` | Always use Runtime, whatever the tags say. |
+| `mantleOnly` | Always use Mantle, whatever the tags say. |
+
+The Kubernetes API takes the same four values capitalized, such as `RuntimePreferred`, under `spec.ai.provider.bedrock.endpointPreference`. A value that you copy from one mode to the other fails to load.
+
+The preference applies to chat completions, messages, responses, and Anthropic token counting. The other route types ignore it: embeddings, reranking, realtime, Gemini token counting, detection, passthrough, and content generation always take Runtime, and model listing always takes Mantle.
```
artberger: I am not sure what are field names here vs generic names. Is embeddings, for example, a field name or just a general name for route type. 

It might help to put field names in code ticks or reword for clarity, maybe something like
```suggestion
The preference applies to chat completions, messages, responses, and Anthropic token counting. The other route types ignore it: embeddings, reranking, realtime, Gemini token counting, detection, and passthrough. The content generation route type always take a Runtime endpoint, and the model listing always takes a Mantle endpoint.
```

assets/agw-docs/snippets/model-catalog-tags.md
kristin-kronstain-brown: Trimmed it down. Kept a short version because the include/exclude lists have to stay a matching pair. Shortening either to `1.5.x` would also match 1.4.x and older, which would show the 1.6 text on versions that don't have the feature. Also trimmed the matching comment in `cost-catalog.md` so the two are consistent.

assets/agw-docs/snippets/model-catalog-tags.md
kristin-kronstain-brown: The rules above do apply, but Bedrock narrows them. Added a `#### Bedrock` heading here so that's visible instead of implied. Kept it under **Chat format tags** rather than moving it down, since the next H3 covers the `runtime`/`mantle` endpoint tags, which are a separate set.

assets/agw-docs/standalone/cost-catalog.md
kristin-kronstain-brown: Good catch, that was ambiguous. Reworded above the table. A source is a catalog to import from, not an LLM provider. Each source covers one or more providers and contributes rates, tags, or both. `models.dev` and `github` cover many providers; `aws-bedrock-mantle` covers Bedrock only and adds tags, not rates.

content/docs/kubernetes/main/documentation/llm/cost-controls/costs.md
kristin-kronstain-brown: Same fix. Added the same source-vs-provider wording above this table.

content/docs/standalone/main/integrations/llm/providers/bedrock.md
kristin-kronstain-brown: It makes sense that the sentence is hard to parse, but I think the suggestion changes the meaning, so what do you think of this?

In `resolve_endpoint`, content generation sits in the same arm as embeddings, reranking, realtime, detection, and passthrough; all of them hard-code Runtime. Splitting it out makes it read like a special case, and the rewrite also drops which endpoint the other six use. I've reworded to make the grouping explicit instead.

> The preference applies to chat completions, messages, responses, and Anthropic token counting. Every other route type ignores it and uses a fixed endpoint. Model listing always uses Mantle, and embeddings, reranking, realtime, Gemini token counting, detection, passthrough, and content generation always use Runtime.

On the code ticks, none of these are field names. They're internal route-type variants that never appear in config, so backticks might suggest they're settable. The only configurable literals in this section are the four `runtimePreferred`-style values, which already have code spans in the table.

content/docs/kubernetes/main/documentation/llm/cost-controls/costs.md
```diff
@@ -32,14 +32,20 @@ Prepare a catalog by creating your own JSON file or using the `agctl catalog imp
...
    ```
 
    > [!IMPORTANT]
-   > The `--providers` flag takes the provider IDs of the source that you import from, and the two sources name some providers differently. The `github` source uses the agentgateway provider IDs, such as `gcp.gemini` and `aws.bedrock`, while `models.dev` uses its own IDs, such as `google` and `amazon-bedrock`. An ID that the `github` source does not recognize is skipped without an error, so check the provider list in the generated file before you load it.
+   > The `--providers` flag takes the provider IDs of the source that you import from, and the sources name some providers differently. The `models.dev` source uses its own IDs, such as `google` and `amazon-bedrock`, while the `github` source uses the agentgateway provider IDs, such as `gcp.gemini` and `aws.bedrock`. The sources also handle an unrecognized ID differently: `models.dev` fails with `no providers matched`, but `github` reports `imported 0 providers` and writes a catalog without that provider. A `--providers` list that omits Bedrock also makes `aws-bedrock-mantle` contribute nothing. Check the provider list in the generated file before you load it.
```
kristin-kronstain-brown:
```suggestion
   > The `--providers` flag takes the provider IDs of the source that you import from, and the sources name some providers differently. The `models.dev` source uses its own IDs, such as `google` and `amazon-bedrock`, while the `github` source uses the agentgateway provider IDs, such as `gcp.gemini` and `aws.bedrock`. The sources also handle an unrecognized ID differently. `models.dev` fails with `no providers matched`, but `github` reports `imported 0 providers` and writes a catalog without that provider. A `--providers` list that omits Bedrock also makes `aws-bedrock-mantle` contribute nothing. Check the provider list in the generated file before you load it.
```

content/docs/standalone/main/integrations/llm/providers/bedrock.md
```diff
@@ -195,8 +203,54 @@ See [here](../anthropic/#use-claude-platform-on-aws) for connect to [Claude Plat
...
+| Value | Endpoint selection |
+|-------|--------------------|
+| `runtimePreferred` | Use Runtime, except for a model tagged `mantle` but not `runtime`. This value is the default. |
+| `mantlePreferred` | Use Mantle, except for a model tagged `runtime` but not `mantle`. |
+| `runtimeOnly` | Always use Runtime, whatever the tags say. |
+| `mantleOnly` | Always use Mantle, whatever the tags say. |
+
+The Kubernetes API takes the same four values capitalized, such as `RuntimePreferred`, under `spec.ai.provider.bedrock.endpointPreference`. A value that you copy from one mode to the other fails to load.
+
+The preference applies to chat completions, messages, responses, and Anthropic token counting. Every other route type ignores it and uses a fixed endpoint: model listing always uses Mantle, and embeddings, reranking, realtime, Gemini token counting, detection, passthrough, and content generation always use Runtime.
```
kristin-kronstain-brown:
```suggestion
The preference applies to chat completions, messages, responses, and Anthropic token counting. Every other route type ignores it and uses a fixed endpoint. Model listing always uses Mantle, and embeddings. Reranking, realtime, Gemini token counting, detection, passthrough, and content generation always use Runtime.
```

artberger (approved):

